PR #47, opened 2023-03-13
Map module using Sumo surfaces

sigurdp:
First cut implementation of Map viewer frontend and backend for Sumo surfaces.
Note that currently there is no actual visualization of the maps - only the png image is shown


anders-kiaer (approved):
🚀

PR #55, opened 2023-03-17
Modify GitHub action workflow to also run tests on branches in main repo

anders-kiaer:


anders-kiaer:
Apparently it now runs - might have been a GHA glitch.

PR #66, opened 2023-03-22
Always cache and compress values in `WorkbenchServices`

sigurdp:
Always use caching in `WorkbenchServices`, and always compress the values (remove duplicates).
We may have to refine this and allow opt-in or opt-out later, but this should serve as a good default behavior.
Also switch to using Javascrip Map instead of Object


rubenthoms (approved):
LGTM :+1:

PR #69, opened 2023-03-24
Use Equinor font

anders-kiaer:
Three-line PR just to feel more at home 🏠 

Also added possibility in dev-mode for `index.html` changes to be reflected immediately without having to rebuild `docker` container.

image


anders-kiaer:
I asked an EDS representative now 🙂 

Their working on a big upgrade of EDS, where they use "style dictionary" to generate platform specific file formats, in that context they might add support for Tailwind as well. There is unfortunately no time estimate on this from EDS. They recommend projects to continue adopting the CDN font `.css` directly.

rubenthoms (approved):
LGTM

PR #73, opened 2023-04-21
Use wsc.SubsurfaceViewer in map module

HansKallekleiv:


.github/workflows/webviz.yml
```diff
@@ -39,6 +39,7 @@ jobs:
       - name: 🏗️ Build JavaScript bundle
         working-directory: ./frontend
         run: |
+          export NODE_OPTIONS="--max_old_space_size=4096"
```
anders-kiaer, line 42: Add a comment why this is needed?

frontend/src/modules/Map/MapView.tsx
```diff
@@ -22,41 +22,48 @@ export function MapView({ moduleContext }: ModuleFCProps<MapState>) {
...
-                <img
-                    alt={surfDataQuery.status}
-                    src={surfDataQuery.data ? `data:image/png;base64,${surfDataQuery.data.base64_encoded_image}` : ""}
-                    style={{ objectFit: "contain", width: wrapperDivSize.width, height: wrapperDivSize.height }}
-                ></img>
-            </div>
         </div>
     );
+            }
+    return (<div/>)
```
anders-kiaer, line 68: Is this correct? I don't see the opening `<div>` this is closing. Maybe it simply can return `null` (nothing)? Maybe even possible to remove the `return` statement?

HansKallekleiv: it works, but it is not correct no. The second return statement should be removed and the first should be wrapped in a "ApiStateWrapper". I couldn't get it to work though as DeckGL raised an exception due to invalid data.
Have to discuss with @rubenthoms

PR #77, opened 2023-04-25
Parameter correlation endpoints

HansKallekleiv:
Initial endpoints for performing parameter/response correlation for timeseries and inplace volumes.
The endpoints are currently missing some arguments, e.g. realization filter and inplace categorical filters.

Attempted to simplify schema types by adding a generic "EnsembleScalarResponse" class for handling the very common case of ending up with a single scalar value for each realization after some calculation.


anders-kiaer (approved):
:+1: We will probably need to refactor the APIs endpoints / names at some point in time, but I think it is good to cover some more use cases before we start on that.

PR #95, opened 2023-04-27
Introduced optional state store persistency

rubenthoms:
Implemented option for storing state store states in local storage. Main use case: Ensemble selection, but other states can also be stored if wanted.


anders-kiaer:
Looks good to me.

One question: How/where in the code should we best ensure that state store is still a valid option? E.g. in the context of ensemble selection, a state stored selected ensemble could be an "invalid" value on next visit (if user does not have read permission to the field in question anymore, or if the ensemble is deleted in the database).

sigurdp (approved):
👍

PR #128, opened 2023-05-10
Log unexpected exceptions to console

anders-kiaer:


anders-kiaer:
I think I'll change from `console.error` to `console.warn` and then takes this in unless someone objects 🙂 It feels more like a warning than an error since we actually catch the error and continue.

image

PR #131, opened 2023-05-11
Avoid re-render of module instances during module dragging

### LINKED ISSUE #82: Unnecessary render
When changing which module is selected/in focus, all modules are rerendered. This also happens when moving/adding a module, even when layout is not changing dynamically (i.e. the layout calculation returns current calculation).

image

anders-kiaer:
It looks like we currently re-render unnecessarily. One source to rerender appears to be `dragPosition` of currently dragged module being given as prop to all instances (also the ones not being dragged). Since layout of module instances is calculated higher up, we should not need to pass this information down to all modules.

Closes #82.

### DIFF
--- a/frontend/src/framework/components/Content/private-components/ViewWrapper/viewWrapper.tsx
+++ b/frontend/src/framework/components/Content/private-components/ViewWrapper/viewWrapper.tsx
@@ -19,11 +19,11 @@ type ViewWrapperProps = {
     height: number;
     x: number;
     y: number;
-    isDragged: boolean;
-    dragPosition: Point;
+    dragPosition: Point | null;
 };
 
-export const ViewWrapper: React.FC<ViewWrapperProps> = (props) => {
+const _ViewWrapper: React.FC<ViewWrapperProps> = (props) => {
+    const isDragged = !!props.dragPosition;
     const [importState, setImportState] = React.useState<ImportState>(ImportState.NotImported);
     const ref = React.useRef<HTMLDivElement>(null);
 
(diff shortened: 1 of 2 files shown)


frontend/src/framework/components/Content/private-components/ViewWrapper/viewWrapper.tsx
```diff
@@ -19,11 +19,11 @@ type ViewWrapperProps = {
     height: number;
     x: number;
     y: number;
-    isDragged: boolean;
-    dragPosition: Point;
```
rubenthoms, line 23: What is the rerender benefit of uniting these two props?

rubenthoms (commented):
I don't think this is sufficient yet for guaranteeing a smooth dragging experience. While this prevents non-active modules from rerendering, the module which is dragged around will still rerender every time the drag position changes. My idea is to have another layer wrapping the module's content which can be memoized while the viewWrapper itself doesn't need to be. If that is not sufficient, we could enable a drag mode in which the content in the view wrapper sub wrapper is hidden.

frontend/src/framework/components/Content/private-components/ViewWrapper/viewWrapper.tsx
anders-kiaer: We can keep them separate wrt. re-rerendering I think 👍

anders-kiaer:
> I don't think this is sufficient yet for guaranteeing a smooth dragging experience. While this prevents non-active modules from rerendering, the module which is dragged around will still rerender every time the drag position changes. My idea is to have another layer wrapping the module's content which can be memoized while the viewWrapper itself doesn't need to be. If that is not sufficient, we could enable a drag mode in which the content in the view wrapper sub wrapper is hidden.

Agree, I didn't know that the content of the dragged module is shown while dragging, but I see know that it is.

PR #132, opened 2023-05-11
Grid3D

HansKallekleiv:


backend/src/backend/primary/main.py
```diff
@@ -28,11 +28,7 @@ def custom_generate_unique_id(route: APIRoute) -> str:
     return f"{route.name}"
 
 
-app = FastAPI(
-    generate_unique_id_function=custom_generate_unique_id,
-    root_path="/api",
-    default_response_class=ORJSONResponse,
-)
+app = FastAPI(generate_unique_id_function=custom_generate_unique_id, root_path="/api")
```
anders-kiaer, line 31: Reverting back from `ORJSONResponse`?

backend/src/backend/primary/routers/grid/router.py
```diff
@@ -0,0 +1,150 @@
...
+    updated_request = Request(
+        scope={
+            "type": "http",
+            "method": request.method,
+            "path": request.url.path,
+            "query_string": request.url.include_query_params(**query_params).query.encode("utf-8"),
+            "headers": request.headers.raw,
+        },
+        receive=request._receive,  # Use the _receive method from the ASGI scope
+    )
```
anders-kiaer, line 73: Are these lines in order to add the query parameters to the `request`? 
 
I guess this will be a useful pattern for other API calls to the user session container. Maybe we should add this logic to the proxy function? Looks like query parameters are available also on the `request` object (if the FastAPI `request` is ~equal `starlette`'s `request` https://www.starlette.io/requests/).

anders-kiaer (changes requested):
Nice to see interactions with grid ensembles being a large step closer 🎉 Some suggestions below from my side.

HansKallekleiv:
Superseded by #136

PR #145, opened 2023-05-22
Added PVT endpoints and prototype module

HansKallekleiv:
To test select case: drogon_ahm_future-2023-05-22

Added a TableAccess for Sumo
- Get descriptions/metadata for all tables in a given realization
- Get a specific realization table in arrow format
- Pseudo code to check if all tables have equal hashes (Not currently available in the Sumo explorer metadata)

Added PVT specific endpoints using the above access

Added frontend module with 1-4 views


backend/src/services/sumo_access/table_access.py
```diff
@@ -0,0 +1,90 @@
...
+
+        case = case_collection[0]
+
+        table_collection = case.tables.filter(
+            name=table_metadata.name,
+            tagname=table_metadata.tagname,
+            iteration=self._iteration_name,
+            realization=realization,
+        )
+        if len(table_collection) == 0:
```
anders-kiaer, line 63: Interesting that `pylint` does not complain on this (https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/use-implicit-booleaness-not-len.html). Does it work with `if not table_collection`?

HansKallekleiv: yes. fixed

anders-kiaer (approved):
👍

PR #152, opened 2023-05-26
Refactor parameters service + Prototype module for visualizing sensitivities

HansKallekleiv:
- Added a Sumo query to get parameter values for an ensemble
- Changed code to expect a flat structure of parameter values instead of a nested structure with optional parameter groups
- Extract sensitivity information from the parameter records and create an `EnsembleParameters` object that contains all relevant information about the parameters
- Add manual aggregation of inplace volumes realization tables (until the Sumo service is matured)

- "Dependent" module for sensitivity analysis. Will expect an input slot with a scalar response for each realization in an ensemble (currently hardcoded to Stoiip). Will have an output slot that returns interactively selected sensitivity.
- Sensitivity calculation for the response is done in Typescript through a SensitivityAccessor class.


backend/src/services/sumo_access/inplace_volumetrics_access.py
```diff
@@ -125,13 +128,50 @@ def get_table(self, table_name: str, column_name: str) -> pa.Table:
...
+            byte_stream: BytesIO = vol_table.blob
+
+            table: pd.DataFrame = pd.read_csv(byte_stream)
+            table["REAL"] = vol_table.realization
+            return table
+
+        with ThreadPoolExecutor() as executor:
+            tables = list(executor.map(worker, list(range(len(vol_table_collection)))))
+            tables = pd.concat(tables)
+            tables = pa.Table.from_pandas(tables)
```
anders-kiaer, line 171: Should we make an issue investigating if `asyncio` could/should be used instead of threads?

anders-kiaer (approved):

PR #166, opened 2023-06-06
Generate production source maps

anders-kiaer:


anders-kiaer:
Still fails during build (too high memory consumption).

PR #183, opened 2023-06-16
Adjusted component structure in order to avoid unnecessary rerenders

### LINKED ISSUE #82: Unnecessary render

rubenthoms:
Closes #82.


HansKallekleiv (approved):
👍

PR #189, opened 2023-06-20
Use `@webviz/subsurface-viewer`

### LINKED ISSUE #96: Map module indefinitely in `loading` state in production bundle
Looks like there is no API calls made from the frontend relevant for the map module.

### LINKED ISSUE #129: Turn on sourcemap
Vite apparently have sourcemaps turned off by default (https://vitejs.dev/config/build-options.html#build-sourcemap).

We should produce the source map in order to help debugging.

As an example, sourcemaps might make it easier to understand e.g. https://github.com/equinor/webviz/issues/96#issuecomment-1542208019

anders-kiaer:
Closes #96 and closes #129.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -13,7 +13,7 @@
                 "@mui/base": "^5.0.0-beta.3",
                 "@tanstack/react-query": "^4.24.10",
                 "@tanstack/react-query-devtools": "^4.24.12",
-                "@webviz/subsurface-components": "^0.4.16-alpha.3",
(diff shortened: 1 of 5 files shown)


anders-kiaer:
I'll merge this in now to get map + grid 3D to work in production + enable source map.

When a new version of `@webviz/subsurface-viewer` is released, with https://github.com/equinor/webviz-subsurface-components/issues/1562 fixed, we will bump the version here and agan use non-default import for `subsurface-viewer`.

PR #197, opened 2023-06-23
Implemented error boundaries, module instance reset and crash view with report function

rubenthoms:
Implemented error boundaries around module views and settings with a crash view and the possibility to reset the crashed instance.

For testing, you can drag in the `InplaceVolumetrics` module and select the ensemble 
`iter-0 (upscaled_grids_realistic_no_unc)`.


anders-kiaer (approved):
🐛 🚀

PR #220, opened 2023-06-28
Smda service

HansKallekleiv:


anders-kiaer:
In order for the CI to get 🟢, you can try adding a dummy variable also for the SMDA key env.variable like done here for `WEBVIZ_CLIENT_SECRET`: 
https://github.com/equinor/webviz/blob/327306242dffa637ddae9ffd8520b955f15ffe4d/.github/workflows/webviz.yml#L60

sigurdp (approved):
Good to go 👍

PR #229, opened 2023-07-06
Implemented `options` parameter for registering modules

rubenthoms:
Instead of having multiple optional parameters which make it cumbersome to only add those really wanted, this PR introduces a new `options` object as input parameter to the `registerModule` method.


sigurdp (approved):
Looks good but a couple of comments:
* Give the options object a separate Typescript type, eg `RegisterModuleOptions`
* Shouldn't `defaultTitle` also be part of the options object?

Not directly related to this PR, but since the file is being touched: Maybe order the modules after importance rather than alphabetically, at least put MyModule1, MyModule2 and DbgWorkspaceSpy at the bottom

sigurdp (approved):
Go for launch 👍

PR #234, opened 2023-07-12
Set `max-warnings` to 0 for eslint

rubenthoms:
We have a lot of warnings for unused imports/variables. Whenever merging into main, we should make sure that all warnings are taken care of. If any warning cannot be fixed, it can be explicitly ignored by using eslint comment commands.


sigurdp (approved):
👍

PR #237, opened 2023-07-12
Added `dependency-cruiser` with custom rules for guaranteeing sane environment

rubenthoms:
Dependency cruiser (https://www.npmjs.com/package/dependency-cruiser, https://snyk.io/advisor/npm-package/dependency-cruiser) is making sure that:

- Files in `src/framework/internal` are not used outside of `src/framework`
- Files in `src/lib` do not import any files from other directories in the source tree (goal is to keep them generic)
- Files in `src/modules/_shared` can only be imported from files within `src/modules`

Added `depscheck` script to `package.json` and included it in `validate` script.

### DIFF
--- a/.github/workflows/webviz.yml
+++ b/.github/workflows/webviz.yml
@@ -44,7 +44,7 @@ jobs:
           export NODE_OPTIONS="--max_old_space_size=4096"
           npm run build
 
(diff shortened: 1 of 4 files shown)


sigurdp (approved):
👍

PR #238, opened 2023-07-12
Housekeeping: prefix private properties and remove unneeded public

sigurdp:
* Prefix all private properties with an `_`
* Removed unneeded use of the `public` visibility modifier
* Fixed some imports so they use aliases instead of relative import paths


rubenthoms (approved):
:+1:

PR #242, opened 2023-07-14
Import `MyModule`, `MyModule2` and `DbgWorkbench` only in dev mode

### LINKED ISSUE #224: Hide/remove "Debug Workbench Spy", "My Module" and "My Module 2"
- "My module" and "My module 2" can probably be removed (later instead replaced by a best practice document for writing modules).
- "Debug workbench spy" maybe is valuable going forward? If that is the case, we could e.g. show it only in dev mode and hide it in production.

Cleaning up by removing modules without value for end users will help getting early testers to understand the framework.

rubenthoms:
Keeping all modules for development/test purposes but not importing in production.

Closes #224.


sigurdp (approved):
👍

PR #251, opened 2023-07-21
Adjustments to slider component

rubenthoms:
Removed internal inputs from slider component (these can be handled from the outside) and added hover value label.

### DIFF
--- a/frontend/src/lib/components/Slider/slider.tsx
+++ b/frontend/src/lib/components/Slider/slider.tsx
@@ -8,13 +8,15 @@ import { BaseComponent } from "../_BaseComponent";
 import { resolveClassNames } from "../_utils/resolveClassNames";
 
 export type SliderProps = {
-    displayValue?: boolean;
+    valueLabelDisplay?: "on" | "auto" | "off";
 } & SliderUnstyledProps;
 
 export const Slider = React.forwardRef((props: SliderProps, ref: React.ForwardedRef<HTMLDivElement>) => {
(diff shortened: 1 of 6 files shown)


frontend/src/lib/components/Slider/slider.tsx
```diff
@@ -135,17 +169,51 @@ export const Slider = React.forwardRef((props: SliderProps, ref: React.Forwarded
...
+                                "before:h-4",
+                                "before:bg-blue-600",
+                                "before:rotate-45",
+                                "before:-z-10",
+                                {
+                                    hidden: !valueLabelVisible,
+                                }
+                            )}
+                        >
+                            {value}
```
jorgenherje, line 212: Utilize prop `valueLabelFormat` if provided?

jorgenherje (changes requested):
Nice update. I have a couple of suggestions for improvement. Otherwise the code looks good to me!

As of now the slider is fed with number or list of numbers for value. Can consider adding possibility of handling additional datatypes, e.g. `string | string[]`, etc. (perhaps `timestep_ms`?)

- Consider adding tick marks for the slider.
- Consider adding possibility for additional datatypes.
- With value as number, allow value label to utilize the prop `valueLabelFormat` s.t. a function can be provided to allow specific label info. Then the user is the responsible to provide function for converting between value index and label string.

jorgenherje (commented):
Looks good to me!

Assuming: Not supporting other data types yet

jorgenherje (approved):

PR #254, opened 2023-08-09
Adjust gitignore for `.vscode`

jorgenherje:
Adjust .gitignore file for vscode settings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,5 +109,4 @@ dist
 
 # .vscode
 .vscode/
-frontend/.vscode/
-backend/.vscode/
+!.vscode/tasks.json
--- a/.gitignore
+++ b/.gitignore
@@ -107,4 +107,7 @@ dist
 # TernJS port file
 .tern-port
 
-frontend/.vscode
+# .vscode
+.vscode/
+frontend/.vscode/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -107,4 +107,7 @@ dist
 # TernJS port file
 .tern-port
 
-frontend/.vscode
+# .vscode
+.vscode/
```
rubenthoms, line 111: This is sufficient since it does exclude all directories named `.vscode` in all subdirectories. It seems there was a tasklist committed to the entry folder's `.vscode` subdir on purpose in order to make the generate api command more accessible from VS Code. Maybe we should not include it in the `.gitignore` list and just add the `backend/.vscode` dir in addition to `frontend`? We can also consider excluding everything except the tasks list in the entry dir.

.gitignore
```diff
@@ -107,4 +107,7 @@ dist
...
+frontend/.vscode/
+backend/.vscode/
```
rubenthoms, line 113: Add a line break at end ;)

rubenthoms (changes requested):

rubenthoms (approved):
:+1:

PR #260, opened 2023-08-18
Add support for historical smry vectors

sigurdp:
Add support for fetching historical vectors from backend

* Implemented backend support for returning historical vector data
* Added fetching and display of historical vector in SimulationTimeSeries module
* Renamed endpoint from "timeseries/vector_names_and_description" to "timeseries/vector_list".
* Cleaned up endpoint parameters and removed unused endpoints/parameters.
* Implemented backend support for removing historical vectors from vector list
* Restructuring of internal code in backend's SummaryAccess class.


HansKallekleiv (approved):
👍

PR #261, opened 2023-08-18
Implemented hook for validated state

rubenthoms:
This new hook does work like the default `React.useState` hook, but in addition to an initial state, it takes a list of valid states and a boolean setting if the state should or should not be updated to a new valid one when the previous one gets invalid. 

The hook checks if the previous state is in the list of valid states. If not, the hook returns the first value from the list of valid states but keeps the original state as it might become valid again. However, if the boolean is set to false, the state is not kept but set to this new valid state.


jorgenherje (approved):
Looks good to me! 

Seems like a hook which can come handy for us!

jorgenherje (approved):
Looks good to me!

PR #264, opened 2023-08-22
Add glob to devDependencies

jorgenherje:
- Add `glob` to devDependecies in package.json
- Update usage from `glob` to `glob.glob` in `compress_static.cjs`


rubenthoms (approved):
:+1:

PR #266, opened 2023-08-23
Adjusted `ResizablePanels` to new React effect recommendations and fixed events for touch use

rubenthoms:
Before this PR, the `ResizablePanels` component could not be properly resized when using a touch device. Moreover, the use of effects was violating React's latest recommendations.


jorgenherje (approved):
Tested and looks good to me! 👍

PR #268, opened 2023-08-25
Add env variable for Sumo environment

HansKallekleiv:
Until we completely move to the Sumo prod environment, we need an env variable to switch between dev and prod.
Added `WEBVIZ_SUMO_ENV` with default set to dev.


rubenthoms (approved):
:+1:

PR #281, opened 2023-08-28
Add export of SmartNodeSelectorSelection

jorgenherje:
Add export of selection type for usage of onChange for both `VectorSelector` and `SmartNodeSelector`


rubenthoms (approved):
:+1:

PR #290, opened 2023-08-31
Persistent ensemble set selection

rubenthoms:
Ensemble set is now stored in local storage and loaded when app is mounted.


sigurdp (approved):
👍

PR #292, opened 2023-08-31
Sensitivity Module - Handle data series with only 0 values

HansKallekleiv:
`Number.MIN_VALUE` is larger than zero causing the axis ranges to be [-Number.MIN_VALUE, Number.MAX_VALUE] which Plotly does not handle.


sigurdp (approved):
Nice 👍 👍

PR #297, opened 2023-09-10
Extended functionality in EnsembleParameters

sigurdp:
* Extended functionality in EnsembleParameters class
* Introduced MinMax class for storing a min/max range

### DIFF
--- a/frontend/src/framework/EnsembleParameters.ts
+++ b/frontend/src/framework/EnsembleParameters.ts
@@ -1,13 +1,65 @@
-export type Parameter = {
+import { MinMax } from "@lib/utils/MinMax";
+
+export enum ParameterType {
+    CONTINUOUS,
+    DISCRETE,
+}
+
+export type ContinuousParameter = {
+    readonly type: ParameterType.CONTINUOUS;
     readonly name: string;
-    readonly isLogarithmic: boolean;    // Only applicable for numerical/float arrays?
-    readonly isNumerical: boolean;
+    readonly groupName: string | null;
+    readonly description?: string;
     readonly isConstant: boolean;
-    readonly groupName?: string;
-    readonly descriptiveName?: string;
-    readonly realizations: number[];       // The two arrays, realizations and values, must always be same length
-    readonly values: number[] | string[];  // Array items can be string, int or float. Should probably be Float32Array, Int32Array or string[] instead.
+    readonly isLogarithmic: boolean;
+    readonly realizations: number[]; // The two arrays, realizations and values, must always be same length
+    readonly values: number[]; // Array items will be floating point.
 };
+
+export type DiscreteParameter = {
+    readonly type: ParameterType.DISCRETE;
+    readonly name: string;
+    readonly groupName: string | null;
+    readonly description?: string;
+    readonly isConstant: boolean;
+    readonly realizations: number[]; // The two arrays, realizations and values, must always be same length
+    readonly values: number[] | string[]; // Array items can be string or int. Should maybe utilize Int32Array or string[] instead?
+};
+
+export type Parameter = ContinuousParameter | DiscreteParameter;
+
+export class ParameterIdent {
+    readonly name: string;
+    readonly groupName: string | null;
+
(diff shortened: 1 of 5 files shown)


frontend/src/framework/EnsembleParameters.ts
```diff
@@ -1,23 +1,124 @@
...
     constructor(parameterArr: Parameter[]) {
         this._parameterArr = parameterArr;
     }
 
-    getParameterNames(): string[] {
-        return this._parameterArr.map((par) => par.name);
+    getParameterIdents(requiredParamType: ParameterType | null): ParameterIdent[] {
+        const identArr: ParameterIdent[] = [];
+        for (const par of this._parameterArr) {
+            if (requiredParamType == null || par.type === requiredParamType) {
```
jorgenherje, line 85: Should it be `===` for `requiredParamType == null`?

frontend/src/framework/EnsembleParameters.ts
```diff
@@ -1,23 +1,124 @@
...
+                identArr.push(new ParameterIdent(par.name, par.groupName));
+            }
+        }
+
+        return identArr;
+    }
+
+    hasParameter(paramIdent: ParameterIdent): boolean {
+        return this.findParameter(paramIdent) != null;
```
jorgenherje, line 94: Should it be `!==`?

frontend/tests/unit-tests/EnsembleParameters.test.ts
```diff
@@ -0,0 +1,118 @@
...
+        {
+            const minMax = ensParams.getContinuousParameterMinMax(ParameterIdent.fromNameAndGroup("cparam_50", "grp1"));
+            expect(minMax).toEqual(new MinMax(51, 54));
+        }
+    });
+});
+
+
+describe("ParameterIdent tests", () => {
+    test("Conversion to/from string", () => {
```
jorgenherje, line 103: Perhaps a matter of taste, but I like to test single functions per test, using hard coded string as expected result for string conversion and hard coded strings as argument for generators from string (`fromString()`method). I.e. test `fromNameAndGroup("aName", "aGroup")`-method towards a `new ParameterIdent("aName", "aGroup")`-object and its public properties? Then test `toString()`-method towards a hard coded expected string? Then testing `fromString()` with the hardcoded string?

jorgenherje (changes requested):
Looks good to me. Nice functionality!

Minor comments

frontend/src/framework/EnsembleParameters.ts
```diff
@@ -1,23 +1,124 @@
...
+
+    static fromString(paramIdentString: string): ParameterIdent {
+        const parts = paramIdentString.split("~@@~");
+        if (parts.length === 1) {
+            return new ParameterIdent(parts[0], null);
+        } else if (parts.length === 2) {
+            return new ParameterIdent(parts[0], parts[1]);
+        } else {
+            throw new Error(`Invalid parameter ident string: ${paramIdentString}`);
+        }
```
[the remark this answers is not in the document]
sigurdp: Agreed 👍

frontend/src/framework/EnsembleParameters.ts
sigurdp: The use of == was intentional and will catch undefined as well. Not a feature win but I see no harm.

frontend/src/framework/EnsembleParameters.ts
sigurdp: Doesn't make much difference, but will change it.

frontend/tests/unit-tests/EnsembleParameters.test.ts
sigurdp: Will make some adjustments to the test

jorgenherje (approved):
👍

frontend/src/framework/EnsembleParameters.ts
jorgenherje: 👍

PR #299, opened 2023-09-11
Fixed/improved multiple components

rubenthoms:
- Added scrollbar to module settings
- Fixed issue with `SmartNodeSelector` tag buttons not being visible and removed `Remove all` button when nothing is selected
- Reduced size of radio buttons, fixed size when long label, removed padding so it's aligned with e.g. checkboxes, added pointer cursor and hover effect
- Increased size of checkbox text
- Added close cross to ensemble dialog
- Improved `useElementBoundingRect` hook to take scrolling into account


jorgenherje (commented):
Looks good to me. Awaiting fix for dropdown suggestions list when scrolling settings area.

PR #301, opened 2023-09-12
Show multiple sensitivities in TimeSeriesSensitivities module

HansKallekleiv:
Option to show more than one sensitivity in TimeSeriesSensitivities module.
Rename module and template.

Awaits #315


jorgenherje (changes requested):
Nice work! Like the functionality!

I would like to get some message in the Tornado Chart when a time step is not selected. It can be difficult to see/know that the `Tornado Chart` awaits a click in the `Simulation Time Series per sensitivity`-module to get an active time stamp. Not sure what is the best way to inform the user of this?

image


Otherwise some minor comments.

frontend/src/modules/SimulationTimeSeriesSensitivity/settings.tsx
```diff
@@ -109,95 +116,91 @@ export function settings({ moduleContext, workbenchSession, workbenchServices }:
...
-                        onChange={handleVectorSelectionChange}
-                        filter={true}
-                        size={5}
-                    />
-                </Label>
+                <CollapsibleGroup expanded={true} title="Time Series">
+                    <Label text="Vector">
+                        <VectorSelector
+                            data={vectorSelectorData}
+                            selectedTags={selectedVectorName ? [selectedVectorName] : []}
```
jorgenherje, line 144: This should not use `selectedVectorName`. It should use a new state named `selectedVectorTag`, which is set in `handleVectorSelectChange` by `selection.selectedTags[0]`. Note that if the a default value or `fixup`-value for `selectedVectorName` is created, this must then also be set to `selectedVectorTag` state.

`selectedVectorName` is used to propagate selected vector to state for view, but `selectedTags` prop for the `VectorSelector` is a tag (e.g. `WOPT:A*`) and not a node.

frontend/src/modules/SimulationTimeSeriesSensitivity/settings.tsx
```diff
@@ -109,95 +116,91 @@ export function settings({ moduleContext, workbenchSession, workbenchServices }:
         console.debug(`handleFrequencySelectionChange()  newFreqStr=${newFreqStr}  newFreq=${newFreq}`);
         setResamplingFrequency(newFreq);
     }
-    if (!sensitivities?.getSensitivityArr()) {
-        return <div>This is not a sensitivity ensemble</div>;
+    function handleVectorSelectChange(selection: SmartNodeSelectorSelection) {
+        setSelectedVectorName(selection.selectedNodes[0]);
```
jorgenherje, line 120: Add `setSelectedVectorTag(selection.selectedTags[0]);` here. See further info below regarding tags vs nodes for `VectorSelector`.

jorgenherje (changes requested):

jorgenherje (approved):

PR #302, opened 2023-09-13
Rename TopographicMap => SubsurfaceMap. Add global colorscale

HansKallekleiv:


frontend/src/modules/SubsurfaceMap/_utils/color.ts
```diff
@@ -0,0 +1,24 @@
...
+    if (match) {
+        return [parseInt(match[1], 10), parseInt(match[2], 10), parseInt(match[3], 10)];
+    }
+    return null;
+}
+export function createContinuousColorScaleForMap(colorScale: ColorScale): colorTablesObj[] {
+    const hexColors = colorScale.getPlotlyColorScale();
+    const rgbArr: [number, number, number, number][] = [];
+    hexColors.forEach((color) => {
+        const rgbString: string = formatRgb(color[1]) as string;
```
rubenthoms, line 17: https://culorijs.org/api/#parse

```typescript
import { parse } from "culori";
...
hexColors.forEach((hexColor) => {
    const rgbColor = parse(hexColor); // Returns object with r, g, b items for hex strings
    rgbArr.push([hexColor[0], rgbColor.r, rgbColor.g, rgbColor.b]);
});
```

rubenthoms (changes requested):
We could consider moving the color scale function into `ColorScale` since it is going to be consumed in several map modules. Otherwise, two comments.

rubenthoms (approved):
LGTM :+1:

PR #304, opened 2023-09-13
Fixed div by 0 issue and assigning max or mid value in these cases

rubenthoms:


jorgenherje (approved):
Looks good to me. Tested in module 👍

PR #307, opened 2023-09-13
Fix wildcard usage for VectorSelector in SimulationTimeSeriesMatrix

jorgenherje:
Use selectedNodes instead of selectedTags to allow usage of wildcard


rubenthoms (approved):
:+1:

PR #309, opened 2023-09-13
Added global error boundary and recovery functionality

rubenthoms:
Also separated Module instance's "importing" from "initialising" loading state.


sigurdp (approved):
👍

PR #310, opened 2023-09-13
Proposal for handling of missing/removed modules

rubenthoms:
This proposal enables the user to recover kindly (keeping other module instances in the layout) when a module in the stored layout no longer exists.


HansKallekleiv (approved):
Very nice 👍 Solves a pain point

jorgenherje (approved):
I really like this feature! When working with multiple branches the layout might contain non-existing modules when switching between these. I would prefer to be able to keep my selected ensembles, and not clear the entire local storage. This PR resolves this need!

This is of course most for development, and not a major need for end user. Removal or renaming of modules might not occur often.

The picture below is a nice feature in my opinion, where I'm able to remove non-existing module and keeping my selected ensembles.

image

PR #311, opened 2023-09-14
Minor refactor of well completion backend and frontend

sigurdp:
Refactored well completion backend and frontend slightly in preparation for introducing new version of FastAPI and Pydantic.
* Removed None as attribute type until we're sure we need it
* Refactored and cleaned up frontend data accessor class in preparation new FastAPI version

There should be no functional changes


jorgenherje (approved):
Looks good to me!

PR #313, opened 2023-09-14
test skeleton for backend with some simple tests

HansKallekleiv:
A first sketch on how to organize tests. 

- Added a folder structure under `tests` with `unit` and `integration`. Subfolders are mirroring the code base. E.g. `integration/services/sumo_access`

- Added one example of unit test and one integration test towards Sumo using interactive token.

- Moved some code out of parameter_access into a utility function.

To run the tests:
```
sumo_login -e dev
pytest
```

- No tests added on the api itself. Should be discussed how/if we want to test the endpoints.
- If we want to test against Sumo we need to handle authentication in github actions and docker-compose. For github actions an option is to use service principal. https://learn.microsoft.com/en-us/azure/developer/github/connect-from-azure?tabs=azure-portal%2Cwindows#use-the-azure-login-action-with-a-service-principal-secret
Example from `fmu-sumo` here: https://github.com/equinor/fmu-sumo/blob/main/.github/workflows/run_tests.yaml

### DIFF
--- a/backend/src/services/sumo_access/parameter_access.py
+++ b/backend/src/services/sumo_access/parameter_access.py
@@ -1,8 +1,6 @@
 import logging
 from typing import List, Optional
 
-import numpy as np
-import pandas as pd
 from fmu.sumo.explorer.explorer import CaseCollection, SumoClient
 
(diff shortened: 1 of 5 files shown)


sigurdp (approved):
👍

PR #316, opened 2023-09-15
Implemented debug profiler wrapper around every module in dev mode

rubenthoms:


jorgenherje (approved):
I like it! Nice feature and useful for testing/development!

PR #318, opened 2023-09-18
Fixes to SmartNodeSelector/VectorSelector

### LINKED ISSUE #315: `VectorSelector` with `maxNumSelectedNodes=1` breaks free text search for vector name
Regression bug after https://github.com/equinor/webviz/pull/308.

It is no longer possible to do free text search for a vector.

rubenthoms:
- Better information when more nodes selected than allowed (via wildcards)
- Reverted disabled wildcards in single-node selection
- Bugfix when blurring selector

Fixes #315.


jorgenherje (commented):
Looks good to me! Await potential fix for discussed bug with backspace issue.

jorgenherje (approved):

PR #340, opened 2023-09-21
Improvements to GUI

rubenthoms:
- Changed icon set to MUI/Google Material Icons (replaced all icons)
- Modules only highlighted when in modules settings/sync settings
- Removed LoginDialog and replaced with StartScreen
- Opening module list when layout is empty (all modules removed)
- Added optional description to modules and added info dialog to modules list items


anders-kiaer (approved):
Tested, works greait. The code changes also looks good to me 👍 

The only thing I'm a bit unsure of if we going forward will need the dialog box for showing module descriptions, or if tooltip on hover is enough (showing only some short formatted description text). Currently I envision tooltip with description text would be enough (and also be an easier/cleaner user interface), but might be wrong.

rubenthoms:
> Tested, works great. The code changes also looks good to me 👍
> 
> The only thing I'm a bit unsure of if we going forward will need the dialog box for showing module descriptions, or if tooltip on hover is enough (showing only some short formatted description text). Currently I envision tooltip with description text would be enough (and also be an easier/cleaner user interface), but might be wrong.

I am also not a big fan of the dialog. However, users with touch devices might not be able to access tooltips. I removed the dialog now for the first implementation. We can decide later on how we are dealing with this.

anders-kiaer (approved):
👍

PR #342, opened 2023-09-21
Added user avatar to `logged_in_user` route and adjusted frontend

### LINKED ISSUE #332: Change to user avatar when logged in
image

At the same time, show real name instead of email-address?

rubenthoms:
Fixes #332.

### DIFF
--- a/backend/src/backend/primary/routers/general.py
+++ b/backend/src/backend/primary/routers/general.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+import json
 
 import starsessions
 from starlette.responses import StreamingResponse
@@ -17,12 +18,10 @@ LOGGER = logging.getLogger(__name__)
 
 class UserInfo(BaseModel):
     username: str
+    display_name: str | None
     avatar_b64str: str | None
     has_sumo_access: bool
     has_smda_access: bool
-
-class UserAvatar(BaseModel):
-    avatar: bytes
 
 router = APIRouter()
 
@@ -54,6 +53,7 @@ async def logged_in_user(request: Request, includeAvatar: bool = Query(False, de
     user_info = UserInfo(
         username=authenticated_user.get_username(),
         avatar_b64str=None,
+        display_name=None,
         has_sumo_access=authenticated_user.has_sumo_access_token(),
         has_smda_access=authenticated_user.has_smda_access_token(),
     )
@@ -64,6 +64,11 @@ async def logged_in_user(request: Request, includeAvatar: bool = Query(False, de
             result = await client.get("https://graph.microsoft.com/v1.0/me/photo/$value", headers=headers)
             if result.status_code == 200:
                 user_info.avatar_b64str = base64.b64encode(result.content)
+        
+        async with httpx.AsyncClient() as client:
+            result = await client.get("https://graph.microsoft.com/v1.0/me", headers=headers)
+            if result.status_code == 200:
+                user_info.display_name = json.loads(result.content.decode("utf-8")).get("displayName", None)
 
     return user_info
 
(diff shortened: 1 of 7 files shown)


backend/src/backend/primary/routers/general.py
```diff
@@ -1,10 +1,14 @@
...
+import json
 
 import starsessions
 from starlette.responses import StreamingResponse
-from fastapi import APIRouter, HTTPException, Request, status, Depends
+from fastapi import APIRouter, HTTPException, Request, status, Depends, Query
 from pydantic import BaseModel
+# Using the same http client as sumo
+import httpx
+import base64
```
anders-kiaer, line 11: Move this to top import group (`base64` is std.lib)?

backend/src/backend/primary/routers/general.py
```diff
@@ -47,12 +52,25 @@ async def logged_in_user(request: Request) -> UserInfo:
...
+        headers = { "Authorization": f"Bearer {authenticated_user.get_graph_access_token()}" }
+        async with httpx.AsyncClient() as client:
+            result = await client.get("https://graph.microsoft.com/v1.0/me/photo/$value", headers=headers)
+            if result.status_code == 200:
+                user_info.avatar_b64str = base64.b64encode(result.content)
+        
+        async with httpx.AsyncClient() as client:
+            result = await client.get("https://graph.microsoft.com/v1.0/me", headers=headers)
+            if result.status_code == 200:
+                user_info.display_name = json.loads(result.content.decode("utf-8")).get("displayName", None)
```
anders-kiaer, line 71: I don't know `httpx` enough yet to immediately know the answer to this: Are these two calls done sequentially here? If they are, and since they are independent, we could perhaps do them in parallell and just `await` until both are done before `returning`? Should speed up the `logged_in_user` response time in that case.

backend/src/services/utils/authenticated_user.py
```diff
@@ -28,6 +30,19 @@ def __eq__(self, other: Any) -> bool:
...
+        if isinstance(self._graph_access_token, str) and len(self._graph_access_token) > 0:
+            return self._graph_access_token
+
+        raise ValueError("User has no graph access token")
+    
+    def has_graph_access_token(self) -> bool:
+        try:
+            self.get_graph_access_token()
+            return True
+        except:
```
anders-kiaer, line 44:
```suggestion
        except ValueError:
```

backend/src/services/utils/authenticated_user.py
```diff
@@ -28,6 +30,19 @@ def __eq__(self, other: Any) -> bool:
 
     def get_username(self) -> str:
         return self._username
+    
+    def get_graph_access_token(self) -> str:
+        if isinstance(self._graph_access_token, str) and len(self._graph_access_token) > 0:
```
anders-kiaer, line 35:
```suggestion
        if isinstance(self._graph_access_token, str) and self._graph_access_token:
```

frontend/src/framework/internal/components/LoginButton/loginButton.tsx
```diff
@@ -29,6 +29,16 @@ export const LoginButton: React.FC<LoginButtonProps> = (props) => {
 
     function makeIcon() {
         if (authState === AuthState.LoggedIn) {
+            if (userInfo?.avatar_b64str) {
+                return (
+                    <img
+                        src={`data:image/png;base64,${userInfo.avatar_b64str}`}
+                        alt="Avatar"
+                        className="w-4 h-4 rounded-full"
```
anders-kiaer, line 37: Should we make the picture slightly larger, same size as icon? 
```suggestion
                        className="w-5 h-5 rounded-full"
```

anders-kiaer (commented):
Nice - this works great! Some small suggestions below, then this is ready for merge 🚀

PR #343, opened 2023-09-22
For discussion only: Temporary changes to make timeseries work with new Sumo cases

HansKallekleiv:
Temporary fixes only as input for discussion.

Can be tested on new cases in sumo "prod" env.
For drogon this case can be used: drogon_sim2sumo-2023-08-30 (one of them...)

Some observations so far:
- schema is missing
- date is in microseconds
- vector columns are in double
- table name is now field specific instead of "summary", i.e. "DROGON"
- tagname is "summary" instead of "eclipse"


HansKallekleiv:
Fixed in: https://github.com/equinor/webviz/pull/434

PR #349, opened 2023-09-25
Fixed two bugs in GUI

rubenthoms:
- Two event handlers not removed in `layout.tsx`
- Stopping event propagation in `header.tsx` on pointer up

---
Visible bugs:
- Jumping to modules list when removing module (although there are still modules left in the layout)
- Removing module by clicking on cross always opened module settings (not wanted when adding/removing modules and having modules list open)


anders-kiaer (approved):
👍🐞

Issue #361, opened 2023-09-27
Missing module previews


anders-kiaer:
Add some more module previews, also for the more advanced ones.

PR #374, opened 2023-09-28
Add more module previews

anders-kiaer:
...for the modules we currently think will keep on living/existing as modules on their own. Closes https://github.com/equinor/webviz/issues/361.


rubenthoms (approved):
LGTM! Nice preview images :+1:

PR #388, opened 2023-10-03
Implemented `GuiMessageBroker`

rubenthoms:
Implemented `GuiMessageBroker` class and added it as a member to `Workbench`. The goal is to remove GUI-related code (`GuiStateStore`) from `Workbench` and also to be able to handle inter-module GUI events in the new class.

**Note:** The custom hook `useGuiState` is making use of React's new `useSyncExternalStore` hook (https://react.dev/reference/react/useSyncExternalStore).


frontend/src/framework/GuiMessageBroker.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    private maybeSavePersistentState(state: GuiState) {
+        if (persistentStates.includes(state)) {
+            // For now, persistent states are only stored in localStorage
+            // However, in the future, we may want to store at least some of them in a database on the server
+            localStorage.setItem(state, JSON.stringify(this._storedValues.get(state)));
+        }
+    }
+
+    subscribeToEvent<K extends GuiEvent>(event: K, callback: (event: GuiEventPayloads[K]) => void) {
```
sigurdp, line 87: Rename `event` to `payload` or `eventPayload`

Any particular reason for using `K` instead of `T` as we usually do? This goes for many of the member functions.
Maybe we should adopt another scheme for naming these parametric types? Noticed that React Query uses names such as `TQueryFnData` and `TError`

frontend/src/framework/GuiMessageBroker.ts
```diff
@@ -0,0 +1,179 @@
...
+        const eventListeners = this._eventListeners.get(event) || new Set();
+        eventListeners.add(callback);
+        this._eventListeners.set(event, eventListeners);
+
+        return () => {
+            eventListeners.delete(callback);
+        };
+    }
+
+    publishEvent<K extends GuiEvent>(event: K, details: GuiEventPayloads[K]) {
```
sigurdp, line 97: Rename `details` to `payload` or `eventPayload`

frontend/src/framework/GuiMessageBroker.ts
```diff
@@ -0,0 +1,179 @@
...
+        moduleName: string;
+        elementPosition: Point;
+        pointerPosition: Point;
+    };
+    [GuiEvent.RemoveModuleInstanceRequest]: {
+        moduleInstanceId: string;
+    };
+};
+
+type GuiStateTypes = {
```
sigurdp, line 42: Should this be more along the lines of `GuiStateValueTypes` ?

sigurdp (approved):
Approved as most of it looks good. A few comments attached.
Should have a short disussion centered around what we call things related to state before merging.
Thinking in terms of `state`, `stateValue`, `stateKey` etc

PR #389, opened 2023-10-03
Update to Python 3.11

anders-kiaer:
Closes https://github.com/equinor/webviz/issues/387.

Also updated `fastapi`, `uvicorn` and `starsessions` to latest `minor/patch` due to exception being thrown in 3.11 from (login?) middleware before updating the dependencies.


HansKallekleiv (approved):
👍

PR #390, opened 2023-10-03
Increase draggable area to whole module preview

### LINKED ISSUE #336: Increase "draggable area" when adding new modules
Currently only the title area (indicated by top yellow rectangle) is "draggable area" when dragging in a new module. Increase this to also include the preview area (bottom drawn yellow rectangle)?

image

anders-kiaer:
Closes #336.

### DIFF
--- a/frontend/src/framework/internal/components/Settings/private-components/modulesList.tsx
+++ b/frontend/src/framework/internal/components/Settings/private-components/modulesList.tsx
@@ -46,12 +46,11 @@ const makeStyle = (isDragged: boolean, dragSize: Size, dragPosition: Point): Rea
 
 const ModulesListItem: React.FC<ModulesListItemProps> = (props) => {
     const ref = React.useRef<HTMLDivElement>(null);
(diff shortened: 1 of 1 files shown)


frontend/src/framework/internal/components/Settings/private-components/modulesList.tsx
```diff
@@ -114,15 +114,15 @@ const ModulesListItem: React.FC<ModulesListItemProps> = (props) => {
             }
         };
 
-        if (ref.current) {
```
rubenthoms, line 117: Line 65 needs adjustment when using `mainRef` instead of `ref`. I suggest removing `ref` and renaming `mainRef` to `ref`.

rubenthoms (changes requested):
LGTM. One change request and one comment. Nice you managed to find and change the respective code so easily. 

Note that the disadvantage with this solution is that the UX now differs from interacting with modules that are already part of the layout where only events captured by the module headers are used to start a move operation.

rubenthoms (approved):
LGTM :rocket:

PR #391, opened 2023-10-03
Moved layout/module handling logic from `Workbench` to new class

rubenthoms:
Name of `LayoutService` class can be discussed.

Depends on #388.


anders-kiaer:
Done

PR #393, opened 2023-10-03
Move config.py to root of package (src)

sigurdp:
Currently `config.py` is being referenced by both `services` and `backend` sub-packages. This breaks the dependency graph since `services` should never import modules from `backend`
* Moved config.py out of backend folder and into top folder of package
* Use SUMO_ENV from config instead of reading environment variable directly in `sumo_access/_helpers`

This change initially broke the tests. Try and remedy this by adding an entry for `pythonpath` under `[tool.pytest.ini_options]` in `poetry.toml` file.


.github/workflows/webviz.yml
```diff
@@ -96,6 +96,9 @@ jobs:
       - name: 🤖 Run tests
         working-directory: ./backend
         run: |
+          export WEBVIZ_CLIENT_SECRET="DUMMY_SECRET"
+          export WEBVIZ_SMDA_SUBSCRIPTION_KEY="DUMMY_SUBSCRIPTION_KEY"
+          export WEBVIZ_SMDA_RESOURCE_SCOPE="DUMMY_RESOURCE_SCOPE"
```
anders-kiaer, line 101: This is solving the same need as we already do in https://github.com/equinor/webviz/blob/389a69f22f1d9f00db08a8c99ec88b00a4399a51/.github/workflows/webviz.yml#L60.

Since we know need them two places, should we define them as `env` variables higher up in the CI workflow? https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow

anders-kiaer (approved):
LGTM. One comment/question only.

.github/workflows/webviz.yml
sigurdp: As long as there is no risk of hiding a requirement (which i don't see) I agree.
Will look into it

PR #404, opened 2023-10-05
Parameter list filtering

jorgenherje:
Component to create a filtered subset of a `Parameter` list. 

Utillizing the `SmartNodeSelector` to filter on the properties of the [Parameter](https://github.com/equinor/webviz/blob/main/frontend/src/framework/EnsembleParameters.ts)-type.

----

Closes: #405


frontend/src/framework/components/ParameterListFilter/private-utils/smartNodeSelectorUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+    createAndAddNode(treeDataNodeList, ParameterParentNodeNames.IS_NONCONSTANT, checkIcon);
+    if (hasContinuousParameter) {
+        createAndAddNode(treeDataNodeList, ParameterParentNodeNames.IS_LOGARITHMIC, checkIcon);
+        createAndAddNode(treeDataNodeList, ParameterParentNodeNames.IS_LINEAR, checkIcon);
+    }
+
+    // Add name and group for parameters
+    for (const parameter of parameters) {
+        addParameterNameAndGroupToTreeDataNodeList(treeDataNodeList, parameter, parentIcon);
+    }
```
rubenthoms, line 88: Consider doing a check and adding the parent nodes once (here) instead of checking for every single parameter if the respective parent node (name/group) needs to be added (removes a lot of unnecessary loops). In addition, are the parameter names unique? If so, you could skip the check for existence for the parameter names and just add them instantly while the group names probably need checking. However, you could also first make a set of group names and then add them.

frontend/src/framework/components/ParameterListFilter/private-utils/smartNodeSelectorUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+    return treeDataNodeList;
+}
+
+export function getParametersMatchingSelectedNodes(parameters: Parameter[], selectedNodes: string[]): Parameter[] {
+    // No selection implies no filtering
+    if (selectedNodes.length === 0) {
+        return parameters;
+    }
+
+    const delimiter = ":";
```
rubenthoms, line 99: Since the delimiter could be changed in the `SmartNodeSelector` component, maybe this should be an argument to the function with a default `:`?

frontend/src/framework/components/ParameterListFilter/private-utils/smartNodeSelectorUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+    GROUP: "Group",
+    CONTINUOUS: "Continuous", // For Parameter.type === ParameterType.CONTINUOUS
+    DISCRETE: "Discrete", // For Parameter.type === ParameterType.DISCRETE
+    IS_CONSTANT: "Constant", // For Parameter.isConstant === true
+    IS_NONCONSTANT: "Nonconstant", // For Parameter.isConstant === false
+    IS_LOGARITHMIC: "Logarithmic", // For Parameter.isLogarithmic === true
+    IS_LINEAR: "Linear", // For Parameter.isLogarithmic === false
+};
+
+export function fromParameterTypeToNodeName(type: ParameterType): string {
```
rubenthoms, line 15: Consider renaming the function to use a verb, maybe `getNodeNameFromParameterType`?

frontend/src/framework/components/ParameterListFilter/private-utils/smartNodeSelectorUtils.ts
```diff
@@ -0,0 +1,179 @@
...
+    if (selectedNodes.length === 0) {
+        return parameters;
+    }
+
+    const delimiter = ":";
+    const findSelectedParameterPropertiesFromName = (propertyName: string): string[] => {
+        return selectedNodes
+            .filter((node) => node.split(delimiter, 1)[0] === propertyName)
+            .map((node) => node.split(delimiter, 2)[1]);
+    };
```
rubenthoms, line 104: Consider moving out of this function and make it a regular function (for debugging). Also, I find the name a bit confusing.

frontend/src/framework/components/ParameterListFilter/parameterListFilter.tsx
```diff
@@ -0,0 +1,83 @@
...
+                return;
+            }
+
+            const filteredParameters = getParametersMatchingSelectedNodes(parameters, selectedNodes);
+            setNumberOfMatchingParameters(filteredParameters.length);
+            if (props.onChange) {
+                props.onChange(filteredParameters);
+            }
+        },
+        [selectedNodes, parameters]
```
rubenthoms, line 58: `props.onChange` needs to be included as dependency.

frontend/src/framework/components/ParameterListFilter/parameterListFilter.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    let candidateTreeDataNodeList = treeDataNodeList;
+    if (parameters === null || !isEqual(props.parameters, parameters)) {
+        candidateTreeDataNodeList = createTreeDataNodeListFromParameters([...props.parameters], checkIcon, segmentIcon);
+        setParameters(props.parameters);
+        setTreeDataNodeList(candidateTreeDataNodeList);
+    }
+    const computedTreeDataNodeList = candidateTreeDataNodeList;
+
+    // Utilizing useEffect to prevent re-render of parent component during rendering
```
rubenthoms, line 41: Consider adding `...during rendering of this component` in order to make it a bit clearer.

frontend/src/framework/components/ParameterListFilter/parameterListFilter.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const [parameters, setParameters] = React.useState<Parameter[] | null>(null);
+    const [treeDataNodeList, setTreeDataNodeList] = React.useState<TreeDataNode[]>([]);
+
+    let candidateTreeDataNodeList = treeDataNodeList;
+    if (parameters === null || !isEqual(props.parameters, parameters)) {
+        candidateTreeDataNodeList = createTreeDataNodeListFromParameters([...props.parameters], checkIcon, segmentIcon);
+        setParameters(props.parameters);
+        setTreeDataNodeList(candidateTreeDataNodeList);
+    }
+    const computedTreeDataNodeList = candidateTreeDataNodeList;
```
rubenthoms, line 39: Consider following the same pattern as suggested in the other PR.

rubenthoms (changes requested):
All in all looks good to me. :+1: There are a few things that need to be fixed and some comments/suggestions.

frontend/src/framework/components/ParameterListFilter/private-utils/smartNodeSelectorUtils.ts
jorgenherje: I considered this, but dropped it as this is private util, and the `SmartNodeSelector` is configured with the default delimiter.
I can add it for clarity, but if so I'll add a `const smartNodeSelectorDelimiter = ":"` in `ParameterListFilter` functional component s.t. the argument is actually used.

frontend/src/framework/components/ParameterListFilter/private-utils/smartNodeSelectorUtils.ts
jorgenherje: 👍

frontend/src/framework/components/ParameterListFilter/private-utils/smartNodeSelectorUtils.ts
rubenthoms: I was only concerned about that somebody else might be editing your code and end up in a situation where things are not working again because they liked to change the delimiter. Makes our code more errorproof. I like your idea to put it into the functional component such that it is passed to both the `SmartNodeSelector` component and the utils function. Then it is very clear.

frontend/src/framework/components/ParameterListFilter/private-utils/smartNodeSelectorUtils.ts
jorgenherje: Done. 

For scoped functionality I sometimes like internal functions. And this is an example where i think arrow-functions come to hand - where it has the `this` content of the parent function, i.e. less arguments.

frontend/src/framework/components/ParameterListFilter/parameterListFilter.tsx
jorgenherje: 👍

frontend/src/framework/components/ParameterListFilter/parameterListFilter.tsx
jorgenherje: ok

frontend/src/framework/components/ParameterListFilter/parameterListFilter.tsx
jorgenherje: This results in infinite loop

rubenthoms: Yes, since we are using an effect here, the consuming component must wrap the function passed to the `onChange` prop inside a `React.useCallback`.

frontend/src/framework/components/ParameterListFilter/parameterListFilter.tsx
jorgenherje: I have made a suggestion and named function `computeTreeDataNodeListAndUpdateStates`. Please review it 👍

frontend/src/framework/components/ParameterListFilter/private-utils/smartNodeSelectorUtils.ts
jorgenherje: I've made a suggestion please review it. Put both parameter name and group name in a set, and then iterate over these to prevent duplicated names/groups. This prevents usage of `findOrCreateNode` method, which is removed.

rubenthoms (approved):
LGTM 🚀

PR #412, opened 2023-10-09
Proposal for renaming/refactoring of functions/classes/files related to Parameters and Sensitivities

rubenthoms:
The current function calls for getting a list of parameters from an ensemble does not really add to readability. 

image

I suggest to (in accordance to implemented patterns around `EnsembleSet`):

- Rename `EnsembleParameters` to `EnsembleParameterSet` (also the related getter function and member)
- Moving `ParameterIdent` into its own file


anders-kiaer:
Closing - framework has changed.

Issue #362, opened 2023-09-27
Communication method between user session container and frontend


anders-kiaer:
Use cases:
- Real time feedback from user session container CPU and memory usage.
- Real time feedback regarding how more inactivity before user session container stops.
- Real time feedback / logging from generic data retrieval / calculations done in user session container

PR #414, opened 2023-10-10
Show live CPU and memory usage of user session backend

anders-kiaer:
Related to #362.

- Replaced the "⭐ GitHub" in the menu with CPU/memory metrics of the user session backend.
- Uses `psutils` in backend to report container stats (https://github.com/giampaolo/psutil/issues/1011#issuecomment-513715809).
- Uses `react-query` on frontend to fetch container status every two seconds. React-query refetches only when user has the browser tab focused. Locking the screen or going away from the tab stopes the refetching, and therefore also starts the countdown for stopping the user session backend.

image

### DIFF
--- a/backend/src/backend/user_session/routers/general.py
+++ b/backend/src/backend/user_session/routers/general.py
@@ -10,11 +10,8 @@ router = APIRouter()
 START_TIME_CONTAINER = datetime.datetime.now()
 
 
-def _convert_psutil_object(psutil_object: NamedTuple) -> Dict[str, Union[str, Dict[str, str]]]:
-    return {
-        key: getattr(psutil_object, key) if key == "percent" else getattr(psutil_object, key)
-        for key in psutil_object._fields
-    }
+def _convert_psutil_object_to_dict(psutil_object: NamedTuple) -> Dict[str, Union[str, Dict[str, str]]]:
+    return {key: getattr(psutil_object, key) for key in psutil_object._fields}
 
 
 @router.get("/user_session_container")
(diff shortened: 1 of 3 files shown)


backend/src/backend/user_session/routers/general.py
```diff
@@ -10,11 +10,9 @@
 START_TIME_CONTAINER = datetime.datetime.now()
 
 
-def human_readable(psutil_object: NamedTuple) -> Dict[str, Union[str, Dict[str, str]]]:
+def _convert_psutil_object(psutil_object: NamedTuple) -> Dict[str, Union[str, Dict[str, str]]]:
```
rubenthoms, line 13: Maybe clarify with `_convert_psutil_object_to_dict`?

backend/src/backend/user_session/routers/general.py
```diff
@@ -10,11 +10,9 @@
...
     return {
-        key: f"{getattr(psutil_object, key):.1f} %"
-        if key == "percent"
-        else f"{getattr(psutil_object, key) / (1024**3):.2f} GiB"
+        key: getattr(psutil_object, key) if key == "percent" else getattr(psutil_object, key)
```
rubenthoms, line 15: I haven't programmed in Python for a while, but this if statement seems to return the same thing no matter what?

frontend/src/framework/internal/components/NavBar/navBar.tsx
rubenthoms, line 1: In the future, I think this is not the perfect place for that info. I think as soon as we have a logging system and even some other data, I would vote for implementing a slim status bar at the bottom of the page. For now, this is good.

rubenthoms (approved):
LGTM! A few comments.

PR #415, opened 2023-10-10
Combine into one Redis instance

### LINKED ISSUE #410: Merge Redis instances
- Prefix namespace.
- redis-user-session

anders-kiaer:
Closes #410.

### DIFF
--- a/backend/src/backend/primary/user_session_proxy.py
+++ b/backend/src/backend/primary/user_session_proxy.py
@@ -1,6 +1,6 @@
 import os
 import asyncio
-from typing import Any
+from typing import Any, Optional
 
 import httpx
 import redis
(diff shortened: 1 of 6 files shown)


backend/src/backend/primary/user_session_proxy.py
```diff
@@ -20,12 +20,20 @@ class RadixJobScheduler:
...
-        self._existing_job_names = redis.Redis(host="redis-user-session-state", port=6380, decode_responses=True)
+
+        # redis.Redis does not yet have namespace support - https://github.com/redis/redis-py/issues/12 - need to prefix manually.
+        self._redis_client = redis.Redis(host="redis-user-session", port=6379, decode_responses=True)
+
+    def _get_job_name(self, user_id: str) -> Optional[str]:
+        return self._redis_client.get("user-job-name-" + user_id)
+
+    def _set_job_name(self, user_id: str, job_name: str) -> None:
+        self._redis_client.set("user-job-name-" + user_id, job_name)
```
sigurdp, line 31: I would suggest extracting the redis access into a small helper class and then utilizing that inside of `RadixJobScheduler`. I feel it is slightly cleaner, *and* it forces all writing/reading to go though a very slim interface.

Something like this:
```
class RedisUserJobs:
    def __init__(self) -> None:
        self._redis_client = redis.Redis.from_url(config.REDIS_URL, decode_responses=True)

    def get_job_name(self, user_id: str) -> Optional[str]:
        return self._redis_client.get("user-job-name:" + user_id)

    def set_job_name(self, user_id: str, job_name: str) -> None:
        self._redis_client.set("user-job-name:" + user_id, job_name)
```

Couple of other points to note:
* Should utilize configuration from `config`
* Don't think there are any hard conventions, but from other code it seems like `:` is often used as a namespace sparator

sigurdp (approved):
Looks good 👍 
Added one comment with suggesed changes in `user_session_proxy.py`

PR #416, opened 2023-10-12
Implemented `StatusController` and `StatusWriter` for setting/updating the status of modules

### LINKED ISSUE #375: Create visualization of query error and fetching status for modules
Make visualization of query error and fetching status for modules

Discussion:
- Visualize in header of module (with icon?)
- On hover or click: Show text/message for error/fetching state

When using multiple useQuery or useQueries (and array of useQuery, i.e. useQuery[]), the modules would benefit of providing info to user if one or more queries has error or are fetching. Some modules might want to visualize the retrieved data and allow further usage if some queries fail, others might depend on successful fetching of all query data.

rubenthoms:
# Description
With this PR, modules can report both their loading status as well as per-render errors/warnings which are displayed in the module's respective header. In addition, custom debug messages can be set which are displayed in the `DebugProfiler` at the bottom of the module's view or settings component, respectively. Moreover, when using this new feature, the module's direct render count (neglecting its sub-components, contrary to the tree render count as viewed by the `DebugProfiler` itself) is reported to the `DebugProfiler` and also displayed there.

Finally, added two new components that ought be used when a module's view or settings component runs into an error or does not have anything to show (`ContentError` or `ContentInfo`, respectively).

---

# How to use the new feature

## In the view component

```typescript
const view = (props: ModuleFCProps<State>) => {
    ...
    const statusWriter = useViewStatusWriter(props.moduleContext);
    ...
    // Report an error that happens while rendering your component
    if (badCondition) {
        statusWriter.addError("My error message");
    }
    ...
    // Report a warning if something went not as expected but the module is still okay
    if (unexpectedButNotBadCondition) {
        statusWriter.addWarning("My warning message");
    }
    ...
    // Set the module's loading state depending on a condition
    statusWriter.setLoading(myQueries.some(query => query.isFetching))
    ...
    // Set a custom debug message
    statusWriter.setDebugMessage("My custom debug message");
    ...
}
```

## In the settings component

*NOTE*: The loading state can currently only be set with the `ViewStatusWriter`.

```typescript
const settings = (props: ModuleFCProps<State>) => {
    ...
    const statusWriter = useSettingsStatusWriter(props.moduleContext);
    ...
    // Report an error that happens while rendering your component
    if (badCondition) {
        statusWriter.addError("My error message");
    }
    ...
    // Report a warning if something went not as expected but the module is still okay
    if (unexpectedButNotBadCondition) {
        statusWriter.addWarning("My warning message");
    }
    ...
    // Set a custom debug message
    statusWriter.setDebugMessage("My custom debug message");
    ...
}
```

---
Closes #375.


frontend/src/framework/StatusWriter.tsx
```diff
@@ -0,0 +1,77 @@
+import React from "react";
+
+import { ModuleContext } from "./ModuleContext";
+import { ModuleInstanceStatusController, StatusMessageType, StatusSource } from "./ModuleInstanceStatusController";
+
+export class ViewStatusWriter {
+    private _statusController: ModuleInstanceStatusController;
+
+    constructor(moduleContext: ModuleContext<any>) {
```
sigurdp, line 9: Maybe just pass the `ModuleInstanceStatusController` directly in the constructor?
Slightly more direct and slightly more cumbersome for the user. After all, we want them to use the hooks.
Same goes for the `SettingsStatusWriter`

frontend/src/framework/StatusWriter.tsx
```diff
@@ -0,0 +1,77 @@
...
+    addWarning(message: string): void {
+        this._statusController.addMessage(StatusSource.Settings, message, StatusMessageType.Warning);
+    }
+
+    setDebugMessage(message: string): void {
+        this._statusController.setDebugMessage(StatusSource.Settings, message);
+    }
+}
+
+function useStatusWriter<
```
sigurdp, line 50: I would prefer to ditch the generics-based `useStatusWriter()` and instead just implement `useViewStatusWriter()` and `useSettingsStatusWriter()` explicitly. 
Results in slight code duplication (but no more code) and will be a win for readability.

frontend/src/framework/ModuleInstanceStatusController.ts
```diff
@@ -0,0 +1,87 @@
...
+export type StatusControllerState = {
+    messages: StatusMessage[];
+    loading: boolean;
+    viewDebugMessage: string;
+    settingsDebugMessage: string;
+    viewRenderCount: number | null;
+    settingsRenderCount: number | null;
+};
+
+export class ModuleInstanceStatusController {
```
sigurdp, line 28: After reading through the code and considering the inheritance and division of labour between `ModuleInstanceStatusController` and `ModuleInstanceStatusControllerPrivate` I think we should approach it in a slightly different way.

I suggest we make `ModuleInstanceStatusController` a pure interface and then defer the entire implementation to a `ModuleInstanceStatusControllerImpl` class. Will be cleaner wrt division of labour and we can lose some of the internal type definitions in `ModuleInstanceStatusController.ts`

Thinking something like this:
```
export interface ModuleInstanceStatusController {
    addMessage(source: StatusSource, message: string, type: StatusMessageType): void;
    clearMessages(source: StatusSource): void;
    setLoading(isLoading: boolean): void;

    setDebugMessage(source: StatusSource, message: string): void;
    incrementReportedComponentRenderCount(source: StatusSource): void;

    reviseAndPublishState(): void;
}
```

frontend/src/framework/StatusWriter.tsx
```diff
@@ -0,0 +1,77 @@
+import React from "react";
```
sigurdp, line 1: File should probably have `.ts` suffix?

sigurdp (commented):
Added some minor suggestions, and some thoughts on an alternative approach to the division of responsibilities between `ModuleInstanceStatusController` and `ModuleInstanceStatusControllerPrivate`

Let's have a discussion before proceeding on this.

jorgenherje (approved):
Tested and I really like the features! With respect to code, Sigurd have written a good feedback there.

Looks good to me, only one question (not needed to be changed)

Question regarding the render of settings error/warnings: These are placed in the header of the module view, is it wanted to have one single place to show these, or could settingsStatusWriter render in the settings header? (I know the settings status will disappear when other module settings or config is rendered in the drawer)

----

Here is a screenshot of testing "invalid" selected vector in settings.

image

sigurdp (approved):
Good to go 👍

PR #419, opened 2023-10-12
Multiple bug fixes and improvements related to initial settings, templates, and `ChannelSelect`

### LINKED ISSUE #306: Issue with initialSettings.applyToStateOnMount
An error is triggered if a module is initially rendered, then replaced with a template using the same module.
i.e. `InitialSettings` is initially empty, then populated by the template which then adds a new hook.

```
Uncaught Error: Rendered more hooks than during the previous render.
```

E.g. to reproduce.
1. Add `Simulation time series sensitivity` module
2. Add `Distribution plot` module
3. Use template `Øyvind's plugin`
4. Observe the error in the console and also that the settings content of the `DistributionPlot` is empty

rubenthoms:
- `Workbench` cannot reset module instance count as it is used as a `key` property for React and would prevent a proper remount of the respective components
- `applyToStateOnMount` needs to be move out of possibly undefined `InitialSettings` and take it as an argument, such that it is always called and the check for `undefined` is moved into the respective `useEffect` (this caused #306)
- `ChannelSelect` did not react to updates to its `initialChannel` property and had a missing dependency in its `useEffect` causing issues when initialising e.g. `TornadoPlot` in a template

Closes #306.

### DIFF
--- a/frontend/src/framework/InitialSettings.ts
+++ b/frontend/src/framework/InitialSettings.ts
@@ -58,17 +58,20 @@ export class InitialSettings {
     isSet(settingName: string, type: keyof InitialSettingsSupportedTypes): boolean {
         return this.get(settingName, type) !== undefined;
     }
+}
 
-    applyToStateOnMount(
-        settingName: string,
-        type: keyof InitialSettingsSupportedTypes,
-        stateSetter: (value: any) => void
-    ): void {
-        React.useEffect(() => {
-            const setting = this.get(settingName, type);
+export function applyToStateOnMount(
+    settingName: string,
+    type: keyof InitialSettingsSupportedTypes,
+    stateSetter: (value: any) => void,
+    initialSettings?: InitialSettings
+): void {
+    React.useEffect(() => {
+        if (initialSettings) {
+            const setting = initialSettings.get(settingName, type);
             if (setting !== undefined) {
                 stateSetter(setting);
             }
-        }, []);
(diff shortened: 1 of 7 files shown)


frontend/src/framework/InitialSettings.ts
```diff
@@ -58,17 +58,20 @@ export class InitialSettings {
...
+}
 
-    applyToStateOnMount(
-        settingName: string,
-        type: keyof InitialSettingsSupportedTypes,
-        stateSetter: (value: any) => void
-    ): void {
-        React.useEffect(() => {
-            const setting = this.get(settingName, type);
+export function applyToStateOnMount(
```
sigurdp, line 63: This function needs a new name since it has lost its naming context by no longer being a member of `InitialSettings` and it doesn't do the application on mount anymore. 
Maybe something along the lines of `maybeApplyInitialSettingsToState()`, `applyInitialSettingsToState()`, `applyNewInitialSettingsToState()` ...

sigurdp (approved):
👍 Looks good. 
Only need to come up with a new name for the `applyToStateOnMount()` function

frontend/src/framework/InitialSettings.ts
rubenthoms: Going with `applyInitialSettingsToState` as `InitialSettings` are only applied once in the lifetime of a module.

PR #427, opened 2023-10-16
Improvements to ensemble selector

### LINKED ISSUE #347: Various new features to ensemble selector
Various improvement suggestions to:
image

First iteration:
- Possibility to filter on cases using same settings as SUMO. For now I think the «keep» toggle is enough on our side.
image
- Possibility to sort list on when case was created
image
- Possibility to filter only cases/iterations upload by logged in user. Would be nice to see the avatar (or username) of user who created the case/avatar in the list. Makes it a bit more personal + you know who to ask if you have questions regarding the iteration/case.
- Possibility to sort cases/iterations both alphabetically and when created. The latter should be default (typically interested in latest case/iteration).

Later iterations:
- Possiblity to define delta ensembles.
- Provide ensemble type metadata to modules (such that they can filter the ensemble list down to only the types giving sense for a module)
- Possibility to rename/define display name for ensembles.
   - Default could be:
      -  ensemble name if unique, if not
      - "case name + ensemble name" if unique, if not
      - "case name + ensemble name (running number)" (this scenario will not happen often I think)

rubenthoms:
Partly fixes #347.

### DIFF
--- a/backend/src/services/graph_access/graph_access.py
+++ b/backend/src/services/graph_access/graph_access.py
@@ -20,18 +20,29 @@ class GraphApiAccess:
             )
             return response
 
-    async def get_user_profile_photo(self) -> str | None:
+    async def get_user_profile_photo(self, user_id: str) -> str | None:
         print("entering get_user_profile_photo")
-        response = await self._request("https://graph.microsoft.com/v1.0/me/photo/$value")
+        request_url = f"https://graph.microsoft.com/v1.0/me/photo/$value"
+        if user_id != "me":
+            request_url = f"https://graph.microsoft.com/v1.0/users/{user_id}/photo/$value"
+
+        print(f"Trying to fetch user photo from: {request_url}")
+        response = await self._request(request_url)
 
         if response.status_code == 200:
             return base64.b64encode(response.content).decode("utf-8")
         else:
+            print(f"Failed ({response.status_code}): {response.content}")
             return None
 
-    async def get_user_info(self) -> Mapping[str, str] | None:
+    async def get_user_info(self, user_id) -> Mapping[str, str] | None:
         print("entering get_user_info")
-        response = await self._request("https://graph.microsoft.com/v1.0/me")
+        request_url = f"https://graph.microsoft.com/v1.0/me"
+        if user_id != "me":
+            request_url = f"https://graph.microsoft.com/v1.0/users/{user_id}"
+
+        print(f"Trying to fetch user info from: {request_url}")
+        response = await self._request(request_url)
 
         if response.status_code == 200:
             return response.json()
--- a/frontend/src/framework/internal/components/SelectEnsemblesDialog/selectEnsemblesDialog.tsx
+++ b/frontend/src/framework/internal/components/SelectEnsemblesDialog/selectEnsemblesDialog.tsx
@@ -2,6 +2,7 @@ import React from "react";
 
 import { CaseInfo_api, EnsembleInfo_api } from "@api";
 import { apiService } from "@framework/ApiService";
+import { useAuthProvider } from "@framework/internal/providers/AuthProvider";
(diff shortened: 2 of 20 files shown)


backend/src/services/graph_access/graph_access.py
```diff
@@ -20,18 +20,26 @@ async def _request(self, url: str) -> httpx.Response:
...
             return response
 
-    async def get_user_profile_photo(self) -> str | None:
-        print("entering get_user_profile_photo")
-        response = await self._request("https://graph.microsoft.com/v1.0/me/photo/$value")
+    async def get_user_profile_photo(self, user_id: str) -> str | None:
+        request_url = f"https://graph.microsoft.com/v1.0/me/photo/$value"
+
+        if user_id != "me":
+            request_url = f"https://graph.microsoft.com/v1.0/users/{user_id}/photo/$value"
```
anders-kiaer, line 27:
```suggestion
        request_url = f"https://graph.microsoft.com/v1.0/me/photo/$value" if user_id == "me" else f"https://graph.microsoft.com/v1.0/users/{user_id}/photo/$value"
```
Subjective, I'm fine as is as well.

backend/src/services/graph_access/graph_access.py
```diff
@@ -20,18 +20,26 @@ async def _request(self, url: str) -> httpx.Response:
...
             return None
 
-    async def get_user_info(self) -> Mapping[str, str] | None:
-        print("entering get_user_info")
-        response = await self._request("https://graph.microsoft.com/v1.0/me")
+    async def get_user_info(self, user_id: str) -> Mapping[str, str] | None:
+        request_url = f"https://graph.microsoft.com/v1.0/me"
+
+        if user_id != "me":
+            request_url = f"https://graph.microsoft.com/v1.0/users/{user_id}"
```
anders-kiaer, line 40: Same her. Consider if they should be treated "equally", or if one of them should be "default" and overridden if not "me".

frontend/src/framework/internal/components/SelectEnsemblesDialog/selectEnsemblesDialog.tsx
```diff
@@ -241,7 +311,7 @@ export const SelectEnsemblesDialog: React.FC<SelectEnsemblesDialogProps> = (prop
                     </div>
                     <div className="flex flex-col flex-grow gap-4 p-4">
                         <Label text="Selected Ensembles">
-                            <table className="w-full border border-collapse table-fixed">
+                            <table className="w-full border border-collapse table-fixed  text-sm">
```
anders-kiaer, line 314:
```suggestion
                            <table className="w-full border border-collapse table-fixed text-sm">
```
When thinking about it, should we introduce https://tailwindcss.com/blog/automatic-class-sorting-with-prettier (in a separate PR/issue)?

frontend/src/framework/internal/components/SelectEnsemblesDialog/selectEnsemblesDialog.tsx
```diff
@@ -198,14 +249,33 @@ export const SelectEnsemblesDialog: React.FC<SelectEnsemblesDialogProps> = (prop
...
+                                    <span className="flex-grow text-sm text-slate-500">
+                                        Select from {caseOpts.length} cases
+                                    </span>
+                                    <Label position="right" text="Keep" title="Show only cases marked as keep">
+                                        <Switch
+                                            checked={casesFilteringOptions.keep}
+                                            onChange={handleKeepCasesSwitchChange}
+                                        />
+                                    </Label>
+                                    <Label position="right" text="My cases" title="Show only cases that I created">
```
anders-kiaer, line 262: I'm fine with would current label title. If we want to align with SUMO (and shorten the text) it would be something like "Show only my cases"
image

anders-kiaer (approved):
Some small comments - otherwise LGTM 👍

PR #428, opened 2023-10-17
Change time series module previews

### LINKED ISSUE #339: Change from dynamically to statically created preview SVGs
Currently we create preview SVGs as React components, e.g. https://github.com/equinor/webviz/blob/main/frontend/src/modules/DistributionPlot/preview.tsx.

Should we instead create preview pictures as static SVGs? 

**Benefits:**
- Less JavaScript code (they are SVG pictures instead).
- Less computation needed by browser on resizing settings tab width (it will instead scale the SVG picture accordingly).
- Easier to make complex SVGs probably (as you do not need to code them, and can instead use e.g. Inkscape to create them interactively).

**Downside:** 
- If we at some point want the content to be dynamic, we will not be able to do that easily since it is not a function creating the SVG anymore.

anders-kiaer:
Closes #339. Changing preview of time series modules

**From:** 
image

**To:**
image


jorgenherje (approved):
Looks good to me! 👍

General comment for preview-visualization for modules: 
I see most of the `.svg`, `.webp` and similar files, provided as `src` for the `` element, are placed at "top level" in the respective module. Should these files/sources for the `` element be placed in a folder, e.g. `private-assets` in the modules?

anders-kiaer:
> Looks good to me! 👍
> 
> General comment for preview-visualization for modules: I see most of the `.svg`, `.webp` and similar files, provided as `src` for the `` element, are placed at "top level" in the respective module. Should these files/sources for the `` element be placed in a folder, e.g. `private-assets` in the modules?

Good question. Separated this into a separate issue, as it also as you say also include the other existing/unchanged modules.

PR #432, opened 2023-10-22
Add observations service/endpoints

HansKallekleiv:


backend/src/services/sumo_access/observation_access.py
```diff
@@ -0,0 +1,70 @@
...
+    async def get_observations(self) -> Observations:
+        """Retrieve all observations found in sumo case"""
+        observations_collection = self._case.dictionaries.filter(
+            stage="case",
+            name="observations",
+            tagname="all",
+        )
+        if len(observations_collection) == 0:
+            return Observations()
+        if len(observations_collection) > 1:
```
anders-kiaer, line 29:
```suggestion
        if await observations_collection.length_async() == 0:
            return Observations()
        if await observations_collection.length_async() > 1:
```

backend/src/services/sumo_access/observation_access.py
```diff
@@ -0,0 +1,70 @@
...
+            raise ValueError(f"More than one observations dictionary found. {observations_collection.names}")
+
+        observations_handle: Dictionary = observations_collection[0]
```
anders-kiaer, line 32:
```suggestion
        observations_handle: Dictionary = await observations_collection.getitem_async(0)
```

backend/src/services/sumo_access/observation_access.py
```diff
@@ -0,0 +1,70 @@
...
+        observations_bytes = observations_handle.blob
+        observations_dict = json.loads(observations_bytes.decode())
```
anders-kiaer, line 34: Some async method that should be used here as well? I do not know what they are called.

PR #434, opened 2023-10-22
Support latest sumo metadata

HansKallekleiv:
- [ ] Use new Sumo parameter table collection for retrieving parameters
- [ ] Use new metadata definition for Sumo Summary data
- [ ] Handle surfaces with missing metadata
- [ ] Set default sumo environment to production


backend/src/services/sumo_access/parameter_access.py
```diff
@@ -20,17 +22,30 @@
...
+            tagname="all",
         )
+        if await table_collection.length_async() == 0:
+            raise ValueError(f"No parameter tables found {self._case.name, self._iteration_name}")
+        if await table_collection.length_async() > 1:
+            raise ValueError(f"Multiple parameter tables found {self._case.name,self._iteration_name}")
 
-        sensitivities = create_ensemble_sensitivities(sumo_parameters)
-        parameters = [create_ensemble_parameter(sumo_parameter) for sumo_parameter in sumo_parameters]
+        table = table_collection[0]
```
anders-kiaer, line 38:
```suggestion
        table = await table_collection.getitem_async(0)
```

backend/src/services/sumo_access/parameter_access.py
```diff
@@ -20,17 +22,30 @@
...
+        byte_stream: BytesIO = table.blob
```
anders-kiaer, line 39: Should this be async in some way?

backend/src/services/sumo_access/summary_access.py
```diff
@@ -302,3 +290,20 @@ def _construct_historical_vector_name(non_historical_vector_name: str) -> Option
...
+        iteration=iteration_name,
+        column=column_name,
+    )
+    table_names = all_smry_table_collections.names
+    if len(table_names) == 0:
+        raise ValueError("No summary table collections found")
+    if len(table_names) == 1:
+        return all_smry_table_collections
+
+    raise ValueError(f"Multiple summary table collections found: {table_names}. Expected only one.")
```
anders-kiaer, line 309: Should this function also be async, and use async/await on its calls to sumo?

anders-kiaer (approved):
A couple of comments - otherwise LGTM.

backend/src/services/sumo_access/parameter_access.py
HansKallekleiv: There is a table.blob_async which do return a BytesIO object, but the following lines:
```python
        table = await table_collection.getitem_async(0)
        byte_stream: BytesIO = await table.blob_async
        table = pq.read_table(byte_stream)
```
gives: `pyarrow.lib.ArrowInvalid: Could not open Parquet input source '<Buffer>': Parquet magic bytes not found in footer. Either the file is corrupted or this is not a parquet file.`

Not entirely sure why...

PR #444, opened 2023-10-25
`Select` and `Virtualization` fixes

### LINKED ISSUE #373: Fix width of `MultiEnsembleSelect`
When settings menu becomes more narrow and the strings/ensemble display names are wider, the selector becomes wider than the settings menu, resulting in horizontal slider.

image

image

### LINKED ISSUE #377: Bug with `Shift+click` in `MultiEnsembleSelect`/ `Select` component
Experiencing issue with `Shift+mouse click` when using `MultiEnsembleSelect` after first render.

**NOTE:** This does not happen after every `ctrl+F5` refresh, so it can be hard to trigger.

-----

Debugging of issue:

* Initially selected ensemble is marked blue. The second is selected while holding `Shift`-button:
image

* Debugging show following in `select.tsx` file: `lastIndex = -1` , resulting in `newSelected = []`?
image

* Resulting in no selected ensembles:
image

### LINKED ISSUE #394: Update scroll position for `Select` component when options are updated
When a selection is made, and the options gets updated the scroll position is kept. This will result in incorrect scroll position, as the selected element might not be placed there anymore. 

When options are updated the scroll position is not set s.t. the selected value is in view.

-----
Here is a test using `SmartNodeSelector` to create a filtered set of `Parameters` as options for the `Select`-component.

1) Select value: 
image

2) Remove `Group` filter to provide longer list of options - the selected value is not in view:
image

3) Scrolling down - the selected value is found:
image

4) Going back to step 1) and updating filter `Group`. Then the selection is not among the new options, thus selected value is updated and set to index 0, but scroll is not updated. 
image

5) Scrolling to the top we find the new selected value:
image

rubenthoms:
- Improved `startIndex` handling in `Virtualization`, fixes #394
- Removed auto `minWidth` handling in `Select`, fixes #373 
- Moved shift handling function out of focus check if statement, fixes #377

### DIFF
--- a/frontend/src/lib/components/Select/select.tsx
+++ b/frontend/src/lib/components/Select/select.tsx
@@ -60,7 +60,6 @@ export const Select = withDefaults<SelectProps>()(defaultProps, (props) => {
     if (!isEqual(filteredOptions, prevFilteredOptions)) {
         let newCurrentIndex = 0;
         let newStartIndex = 0;
-        const indexOffset = currentIndex - startIndex;
         let oldCurrentValue = prevFilteredOptions[currentIndex]?.value;
         if (props.value?.length >= 1) {
             oldCurrentValue = props.value[0];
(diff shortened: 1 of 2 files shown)


jorgenherje (changes requested):
Codewise looks good. Tested and verified text overflow in select component and shift+click on first render.

As discussed, it seems to an issue with scroll position when updating content of select component.

jorgenherje (approved):
Looks good to me! 👍

PR #453, opened 2023-11-01
Remove hardcoded Drogon field/case selection

### LINKED ISSUE #452: Do not require Drogon access
Currently Drogon is injected into the fields list and authorization fails if a user does not have access to Drogon.

HansKallekleiv:
Closes #452


anders-kiaer (approved):
👍

PR #457, opened 2023-11-03
Use case insensitive string comparison when filtering on user

### LINKED ISSUE #455: Case disappear when My cases is toggled on
In the "Select ensembles" dialogue box, my case disappear when I toggle ON "My cases":
image
image

HansKallekleiv:
Fixes #455


rubenthoms (approved):
:+1:

PR #458, opened 2023-11-03
Improved loading UX

rubenthoms:
- Changed UX in ensemble select dialog -> dialog closes when ensembles are loaded
- Show loading screen while working session is restored (i.e. ensembles are loaded)
- Moved ensemble loading state to `WorkingSessionPrivate`


anders-kiaer (approved):
Tested also in Radix - LGTM 👍🚀

PR #459, opened 2023-11-04
Parameter Distribution Matrix

HansKallekleiv:
Module to visualize many parameter distributions fast.
For single ensembles, ahm parameter updates or other ensembles that have overlapping parameters.

image


anders-kiaer (approved):
I like it 👍🚀 Also a nice example of new standalone module without new APIs.

Issue #448, opened 2023-10-26
Add observations to `SimulationTimeSeriesMatrix` module


jorgenherje:
Add observations for time series, as end points are added.

PR #462, opened 2023-11-06
Add observations to SimulationTimeSeriesMatrix

jorgenherje:
Add observations to plot

---

Closes: #448


anders-kiaer (approved):
Nice 👍 Probably a separate issue: Should we (and if so how) inform the user if there is no summary observations for selected ensemble(s)?

jorgenherje:
> Nice 👍 Probably a separate issue: Should we (and if so how) inform the user if there is no summary observations for selected ensemble(s)?

Good point. Perhaps we could provide it as a warning text to the `statusWriter`? If so, is it of interest to have it per selected vector or for the selected ensemble?

anders-kiaer:
Most of the vectors in an ensemble with summary observations will still not have observations (i.e. there are only a few of the vectors in the set of vectors which potentially will have observations), so in order to avoid warning fatigue maybe try ensemble level?

jorgenherje:
I've rewritten the query code to easier utilize data info and to show status when an ensemble does not have observations.

Made use of the new `combine`, where I see if the ensemble query gets any summary observations. If empty list from back-end is retrieved, the ensemble does not have observations.

HansKallekleiv (approved):
👍

frontend/src/modules/SimulationTimeSeriesMatrix/utils/subplotBuilder.ts
```diff
@@ -466,8 +472,36 @@ export class SubplotBuilder {
         }
     }
 
-    addVectorObservations(): void {
-        throw new Error("Method not implemented.");
+    addVectorObservations(
+        vectorsObservationData: {
```
rubenthoms, line 476: `vectorObservationData`?

rubenthoms (approved):
LGTM :+1: Two small comments.

PR #463, opened 2023-11-06
Enabled debugging for Python backend

rubenthoms:
Enables debugging of backend code via VS Code. Just start the application as usual and then go to `Run and Debug` in VS Code. Run `Python: Remote Attach` and start setting breakpoints. Remember to disconnect in VS Code first when wanting to shut down the docker containers, otherwise, the backend primary container cannot be stopped gently and has to be forced to stop. There is still a small issue: When you close the debugging connection, you cannot attach to the Python process again and you have to restart the Docker container.


jorgenherje (approved):
Looks good! Very nice feature!

HansKallekleiv (approved):
Nice feature 👍

PR #467, opened 2023-11-07
Allow change of reference sensitivity in TornadoChart/Table

### LINKED ISSUE #465: `rms_seed` required
**How to reproduce:**

1) Add _Simulation time series sensitivity_
2) Add tornado chart.
3) Attach the two using data channels.

This will give as error message in console:
```
view.tsx:107  Error: SensitivityResponseCalculator: Reference sensitivity rms_seed not found in ensemble
    at new ze (sensitivityResponseCalculator.ts:57:19)
    at Oe (view.tsx:104:51)
    at G1 (react-dom.production.min.js:167:137)
    at pg (react-dom.production.min.js:197:258)
    at JC (react-dom.production.min.js:292:88)
    at QC (react-dom.production.min.js:280:389)
    at m8 (react-dom.production.min.js:280:320)
    at qf (react-dom.production.min.js:280:180)
    at WC (react-dom.production.min.js:268:209)
    at k (scheduler.production.min.js:13:203)
```

**The user only see in frontend that there is no data provided through data channel:** 

image

**Possible solutions include:**

Instead of requiring `rms_seed`, let user choose reference seed in tornado chart module. Could let `*_seed` be default if it exists, like now.

HansKallekleiv:
Tries to set to set `rms_seed` as reference, if not picks the first sensitivity found.
The reference can then be changed from a dropdown in the TornadoChart module.

Closes #465

### DIFF
--- a/frontend/src/modules/SimulationTimeSeriesSensitivity/settings.tsx
+++ b/frontend/src/modules/SimulationTimeSeriesSensitivity/settings.tsx
@@ -96,7 +96,7 @@ export function settings({ moduleContext, workbenchSession, workbenchServices }:
                 setSelectedSensitivities(sensitivityNames);
             }
         },
-        [sensitivityNames]
+        [computedEnsemble]
     );
     const sensitivityOptions: SelectOption[] = sensitivityNames.map((name) => ({
         value: name,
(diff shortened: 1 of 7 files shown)


anders-kiaer (approved):
LGTM 🐞👍

Issue #440, opened 2023-10-24
Expand/fix usage of `async` for SUMO access


sigurdp:
PR #378 incorporates usage of `async` in many/most of the access layer functions, but we should revisit this to ensure that we use `async` calls all the way where possible. 

Some exampler of functions that should be refactored are:
* `SurfaceAccess.get_realization_surface_data()`
* `SurfaceAccess.get_statistical_surface_data()`
* `SurfacePolygonsAccess.get_surface_polygons()`

Some of these are non-async now, while others are defined as async, but do not actually use underlying async calls to Sumo.

Another related topic is whether we should have dual (both sync and async) functions for all or some of these access functions.
While seeking performance optimizations, we have done several experiments involving Python `multiprocessing`, and in that case, having async functions quickly becomes a nuisance, since by default there is no event loop running.

In that case, how should we name the functions?
We may for example need both:
`SurfaceAccess.get_realization_surface_data_sync()` and `SurfaceAccess.get_realization_surface_data_async()`

as well as both:
`SurfaceAccess.from_case_uuid_sync()` and `SurfaceAccess.from_case_uuid_async()`

etc.

anders-kiaer:
Mentioned functions refactored into async, and with `_async` suffix. Naming convention introduced.

PR #475, opened 2023-11-11
Make functions in `SurfaceAccess` truly async

sigurdp:
- Made `get_realization_surface_data()` and `get_statistical_surface_data()` truly async
- Updated routes that consume these accodingly
- Started establising convention with postfixing async functions with `_async`

Addresses some of the issues raised in #440


HansKallekleiv (approved):
👍

PR #298, opened 2023-09-11
Refactor surface directory

HansKallekleiv:
- [x] Replace existing `StaticSurfaceDirectory` and `DynamicSurfaceDirectory` endpoints and service layer with a single `SurfaceDirectory`, which is a list of  all surfaces with additional metadata sorted by stratigraphical order. Non-stratigraphical surfaces are placed last in the list.
- [x] Rename surface metadata  to more descriptive properties.
- [x] Move some of the logic of filtering the directory to a typescript class

- ~~[ ] Add reusable `SurfaceSelect` react component~~

- [x] Update existing `Map` and `TopographicMap` modules to use the new directory
- [x] Add stratigraphy ordering from Smda.
- [x] Update remaining endpoints and logic to retrieve the surface data

Closes #265

PR #476, opened 2023-11-12
Add polygons directory

HansKallekleiv:
Add/update polygons functionality similar to SurfaceDirectory in https://github.com/equinor/webviz/pull/298

Needed for upcoming https://github.com/equinor/webviz/pull/472

Surface and polygons share alot of metadata, and uses same logic for e.g. validating against official stratigraphy,
so this can be further simplified later on.

There are also helper functions buried in the modules that can be shared. Keeping it as is now until things mature more.


sigurdp (approved):
👍

PR #482, opened 2023-11-23
Bump sumo

HansKallekleiv:
Bump sumo to implement this fix: https://github.com/equinor/fmu-sumo/issues/261
Some small changes required to es queries due to api changes in sumo-wrapper.

PR #486, opened 2023-11-30
Fix Sumo masterdata queries

HansKallekleiv:
Regression fix after https://github.com/equinor/webviz/pull/482 due to underlying change of api in https://github.com/equinor/sumo-wrapper-python/pull/154

- Changing masterdata queries to use post for consistency with the other queries.


sigurdp (approved):
👍

Issue #447, opened 2023-10-26
Prevent disable of `VectorSelector` if one of the selected ensembles has vectors!


jorgenherje:
If one or more of the selected ensembles does not have vectors, the ApiStateWrapper prevents interaction with `VectorSelector`. Prevent disable, and rather keep just warning message in view as is.

See:

![image](https://github.com/equinor/webviz/assets/82032112/2d77ad4c-4f6e-42de-afaa-5b54b9ccf834)

Issue #417, opened 2023-10-12
Refactor `ApiStateWrapper` and `ApiStatesWrapper`


jorgenherje:
Refactor code to have one single wrapper taking either a single apiResult or a list of apiResults.

Replace usage of `isLoading` to `isFetching` to obtain correct "loading" state when using react-query.

PR #493, opened 2023-12-04
Replacing ApiStateWrapper and ApiStatesWrapper with QueryStateWrapper

jorgenherje:
Replace state wrappers `ApiStateWrapper` and `ApiStatesWrapper` with a single renamed wrapper handling both single query and list of queries as prop.

For list of queries, it is possible to configure when to display error state: 
1, When all queries has error
2. When at least one query has error.

---
Closes: #417
Closes: #447


frontend/src/lib/components/QueryStateWrapper/queryStateWrapper.tsx
```diff
@@ -0,0 +1,73 @@
...
+    props: QueryStateWrapperProps | QueryStatesWrapperProps
+) => {
+    let showQueryLoading = false;
+    let showQueryError = false;
+    if ("queryResult" in props) {
+        showQueryLoading = props.queryResult.isLoading;
+        showQueryError = props.queryResult.isError;
+    } else {
+        showQueryLoading = props.queryResults.some((elm) => elm.isLoading);
+        if (props.queryResults.length > 0) {
```
rubenthoms, line 42: This check to prevent `every` from returning true on empty arrays could be moved to line 41.

frontend/src/modules/Grid3DIntersection/settings.tsx
```diff
@@ -65,39 +67,37 @@ export function settings({ moduleContext, workbenchSession }: ModuleFCProps<stat
...
                     />
                 </Label>
 
-                <Label
-                    text="Realizations"
-                >
+                <Label text="Realizations">
                     <Select
-                        options={stringToOptions(allRealizations as any || [])}
+                        options={stringToOptions((allRealizations as any) || [])}
```
rubenthoms, line 82: Why this typechange to any? `allRealizations` is set to be of type `string[]`, no transforming required here. I know it's not your code but since you are already changing it here, we could remove this.

rubenthoms (approved):
Some suggestions, otherwise LGTM.

frontend/src/lib/components/QueryStateWrapper/queryStateWrapper.tsx
```diff
@@ -0,0 +1,73 @@
...
+export const QueryStateWrapper: React.FC<QueryStateWrapperProps | QueryStatesWrapperProps> = (
+    props: QueryStateWrapperProps | QueryStatesWrapperProps
+) => {
+    let showQueryLoading = false;
+    let showQueryError = false;
+    if ("queryResult" in props) {
+        showQueryLoading = props.queryResult.isLoading;
+        showQueryError = props.queryResult.isError;
+    } else {
+        showQueryLoading = props.queryResults.some((elm) => elm.isLoading);
```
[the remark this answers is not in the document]
jorgenherje: True, `isFetching` is correct.

Issue #478, opened 2023-11-15
WellCompletions: Consider replacing hardcoded stratigraphy colors in back-end with colorpalette front-end


jorgenherje:
Consider:

As a first iteration to prevent hardcoded drogon stratigraphy colors back-end, make usage of discrete color palette front-end and iterate over the palette per zone/subzone when creating the well completions plot.

Assign a color per subzone, and thereafter filter? Or assign after filtering. The first ensures consistent colors, the second might end up duplicating colors if filtering is applied.

PR #495, opened 2023-12-04
Use colorpalette for stratigraphy colors in front-end for `WellCompletions`-module

jorgenherje:
- Assign stratigraphy colors using color palette
- Add usage of status writer for query errors/status

----

Closes: #478


anders-kiaer (approved):
👍

PR #508, opened 2023-12-12
Implementation of `RealizationPicker` component

rubenthoms:
Added a new component for picking realizations in a more user-friendly way than a plain input field. This component is a basic version and might require additional updates/features in order to align with the requirements of further use cases.


jorgenherje (approved):
👍

PR #509, opened 2023-12-12
Adjustments to `useValidState` hook

rubenthoms:
Supports a validating function as an argument now and has a more intuitive interface.
Also implements a new hook for dealing with array states.


jorgenherje (approved):
I like the fixes, nice features!

Tested usage of the hooks, and it looks good to me! 👍

PR #510, opened 2023-12-13
New `eslint` rules for React and adjustments and bug fixes to codebase

rubenthoms:
This PR takes in use the recommended rules for `react` and `react-hooks`. It also implements the required changes to the codebase, e.g. missing dependencies in `useEffect`, non-standard-conform hook and component names, non-reference-stable state setters (e.g. in `StateStore`), and possibly changed references in unmount functions. It does also fix a bug in `TableSelect` which led to a jumping behaviour of the scrollbar.

Note: According to the React documentation, reference-stable variables (like state setters) do not need to be passed to dependency arrays. However, the `rules-of-hooks` plugin is not adjusted yet and requires this. In order to benefit from its other bug identification features/advantages, we are now adjusting to its requirements.

### DIFF
--- a/frontend/src/modules/WellCompletions/settings.tsx
+++ b/frontend/src/modules/WellCompletions/settings.tsx
@@ -27,7 +27,7 @@ enum RealizationSelection {
     Single = "Single",
 }
 
-export const settings = ({ moduleContext, workbenchSession, workbenchServices }: ModuleFCProps<State>) => {
+export const Settings = ({ moduleContext, workbenchSession, workbenchServices }: ModuleFCProps<State>) => {
     const ensembleSet = useEnsembleSet(workbenchSession);
     const [availableTimeSteps, setAvailableTimeSteps] = moduleContext.useStoreState("availableTimeSteps");
     const setDataLoadingStatus = moduleContext.useSetStoreValue("dataLoadingStatus");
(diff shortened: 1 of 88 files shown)


frontend/src/modules/WellCompletions/settings.tsx
```diff
@@ -116,7 +147,14 @@ export const settings = ({ moduleContext, workbenchSession, workbenchServices }:
             }
             createAndSetPlotData(allTimeSteps, timeStepIndex, selectedTimeStepOptions.timeAggregationType);
         },
-        [wellCompletionsQuery.data, selectedTimeStepOptions]
+        [
+            wellCompletionsQuery.data,
+            selectedTimeStepOptions,
+            availableTimeSteps,
```
jorgenherje, line 153: Wouldn't this re-trigger the useEffect when line 109 and 119 setAvailableTimeSteps?

jorgenherje (changes requested):
Seems good, but I have a couple of comments with respect to:

- Empty dependency array for `React.useEffect`, which is intended as initialization done on the initial component render.
- Some comments are not directly a result of this PR, but comments are added as I saw them now. E.g. if `ContentInfo`/`ContentError` component can be used.

jorgenherje (approved):
Looks good to me! 👍

PR #515, opened 2023-12-18
Add dev label

### LINKED ISSUE #512: Add information stating that Webviz is under development
There are 5+ fields testing new Webviz today, and the number of users on each of these fields is increasing steadily.

It is easy for the (test) users to forget that the Webviz version they interact with is under heavy development. Should add a clear reminder (e.g. on login page?) and/or under the Webviz logo in the app itself that 
- the application is under heavy development
- expect errors and issues
- We are very interested in feedback on e.g. Slack (https://equinor.slack.com/messages/webviz/) or Viva Engage / Yammer (https://web.yammer.com/main/groups/eyJfdHlwZSI6Ikdyb3VwIiwiaWQiOiIxMzM5NzE0NyJ9)

rubenthoms:
Closes #512.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -15,6 +15,19 @@ import { useQueryClient } from "@tanstack/react-query";
 import "./modules/registerAllModules";
 import "./templates/registerAllTemplates";
 
+function DevLabel() {
+    return (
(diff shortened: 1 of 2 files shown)


anders-kiaer (approved):
LGTM!

I think we might loose some feedback if we do not provide other feedback channels than GitHub. Most of our users are non-technical without GitHub knowledge or user account, but they know about especially Yammer, but also Slack.

PR #521, opened 2023-12-22
Adjust warning message for statusWriter in SimulationTimeSeriesMatrix module

jorgenherje:
Correct/fix warning messages to show correct content

- Ensure warning message with non existing vectors when selecting single ensemble without vectors.
- Only show message for non-existing vector observations when `showObservations` is selected/true


anders-kiaer (approved):
👍

PR #504, opened 2023-12-08
Refactored data channels

rubenthoms:
This PR contains the following changes:
- Refactoring of data channels (naming and concept)
- Performance improvements (related to data channel experience)
- Refactoring of consumers of data channels (publishers and receivers)

Closes #421.
Closes #184 although there might be some further improvements.

Some performance improvements were taken out of this PR due to unknown possible consequences and are going to be implemented in another PR.

PR #532, opened 2024-01-12
Performance improvements

rubenthoms:
Overall performance improvements that include:
- Removed `select-none` class in `Layout` and `SmartNodeSelector`
- Added extra root element for React portals and other global elements (in order to prevent repainting of the main root) - implemented custom component for adding portals
- Made event listeners conditional whenever possible (e.g. event listeners for `pointermove` and `pointerup` for dragging are now added in the `pointerdown` handler and removed again after the dragging operation has finished)
- Improved the `useElementSize` and `useElementBoundingRect` hooks in order to prevent (potentially expensive) re-rendering when an element becomes invisible and visible again without actually changing its size
- Improved performance of the MutationObserver in `useElementBoundingRect`
- Limited max number of vectors in `SimulationTimeSeriesMatrix` to 20 and set the default display mode to `Statistical fanchart`
- Improved text width calculation by preventing it from repeatedly changing the DOM or expensively calculating styles (fonts are now calculated once  for 1rem font-size, stored and later used for calculating text widths)
- Improved layout dragging performance by hiding and not changing the size of modules but using empty preview containers - modules are resized when the dragging operation ends
- Improved layout dragging UX by calculating the dragging point relative to the module's width - if the module width changes, the dragging point (point where the cursor is positioned relative to the module) is changed accordingly
- Replaced arrow functions with proper functions
- Replaced heavy event handlers with smarter logic
- Added event handlers for `blur` and `pointercancel`
- Added `PendingWrapper` component for usage with React's `useTransition` hook
- Made use of `useTransition` in `SimulationTimeSeriesMatrix` and improved settings render code by moving a heavier computing function into the handler function that actually updates its inputs
- Improvements to `SmartNodeSelector`

Note: There are already some performance improvements implemented in #504 which, together with these improvements, should help to improve the overall UX. This PR should not be merged before #504 is.


frontend/src/lib/components/Select/select.tsx
```diff
@@ -48,16 +48,24 @@ export const Select = withDefaults<SelectProps>()(defaultProps, (props) => {
...
     const [currentIndex, setCurrentIndex] = React.useState<number>(0);
+    const [options, setOptions] = React.useState<SelectOption[]>(props.options);
+    const [prevOptions, setPrevOptions] = React.useState<SelectOption[]>(props.options);
     const [prevFilteredOptions, setPrevFilteredOptions] = React.useState<SelectOption[]>([]);
 
     const ref = React.useRef<HTMLDivElement>(null);
     const noOptionsText = props.placeholder ?? "No options";
+
+    if (!isEqual(props.options, prevOptions)) {
+        setPrevOptions(props.options);
```
jorgenherje, line 59: Is it need for `prevOptions`, as both `options` and `prevOptions` states are set simultaneously?

NB: on line 231 below, the `props.options` are used in the returned component render code. Should it be using `options` state instead?

frontend/src/lib/utils/geometry.ts
```diff
@@ -18,145 +18,161 @@ export type Rect = {
...
 export const MANHATTAN_LENGTH = 13.11;
 
-export const pointerEventToPoint = (event: PointerEvent): Point => {
+export function pointerEventToPoint(event: PointerEvent): Point {
     return { x: event.pageX, y: event.pageY };
-};
+}
 
-export const vectorLength = (vector: Point): number => {
+export function vectorLength(vector: Point): number {
```
jorgenherje, line 25: Should `Vector2D` be a new type? It would of course be a type with `x: number` and `y: number`, but for clarity it is not a `Point` with x- and y-coordinate, but a length/size in x and y.

frontend/src/lib/utils/geometry.ts
```diff
@@ -18,145 +18,161 @@ export type Rect = {
...
+}
 
-export const pointMultiplyWithScalar = (point: Point, scalar: number): Point => {
+export function pointMultiplyWithScalar(point: Point, scalar: number): Point {
     return { x: point.x * scalar, y: point.y * scalar };
-};
+}
 
-export const isPartlyContained = (
+export function pointMultiplyWithVector(point: Point, vector: Point): Point {
```
jorgenherje, line 57: Same comment as above?

frontend/src/lib/utils/textSize.ts
```diff
@@ -1,12 +1,38 @@
-export function getTextWidthWithFont(text: string, font: string): number {
-    // re-use canvas object for better performance
+type CharWidths = Map<string, number>;
+
+const text = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_.!~*'()[]{}/?:@#$%^&*+=`|\\\"<>";
+const storedFontWidths: Map<string, CharWidths> = new Map();
+
+function calcAndStoreCharWidths(font: string) {
```
jorgenherje, line 6: Consider adding return type for readability

frontend/src/lib/utils/textSize.ts
```diff
@@ -1,12 +1,38 @@
-export function getTextWidthWithFont(text: string, font: string): number {
-    // re-use canvas object for better performance
+type CharWidths = Map<string, number>;
+
+const text = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_.!~*'()[]{}/?:@#$%^&*+=`|\\\"<>";
```
jorgenherje, line 3: Consider renaming to `textChars`, `validChars`, etc, to deviate from the `text` argument of `getTextWidthWithFont`-function below.

frontend/src/modules/DistributionPlot/view.tsx
```diff
@@ -17,9 +19,20 @@ import { makeHistogramTrace } from "./utils/histogram";
 import { makeHoverText, makeHoverTextWithColor, makeTitleFromChannelContent } from "./utils/stringUtils";
 import { calcTextSize } from "./utils/textSize";
 
+const MaxNumberPlotsExceededMessage: React.FC = () => {
+    return (
+        <ContentWarning>
+            <Warning fontSize="large" className="mb-2" />
+            Too many plots to display. Due to performance limitations, the number of plots is limited to 12.
```
jorgenherje, line 26: Make: `const MAX_NUMBER_OF_PLOTS = 12;`
Make the number a prop for the `MaxNumberPlotsExceededMessage`-component and use the `MAX_NUMBER_OF_PLOTS` for the checks below.

frontend/src/modules/DistributionPlot/view.tsx
```diff
@@ -135,6 +148,10 @@ export const View = ({ moduleContext, workbenchSettings }: ModuleFCProps<State>)
 
             if (plotType === PlotType.Histogram) {
                 const numContents = receiverX.channel.contents.length;
+                if (numContents > 12) {
```
jorgenherje, line 151: Use const max here?

frontend/src/modules/DistributionPlot/view.tsx
```diff
@@ -184,12 +201,16 @@ export const View = ({ moduleContext, workbenchSettings }: ModuleFCProps<State>)
...
                     }
                 }
-                setPlot(figure.makePlot());
+                setContent(figure.makePlot());
                 return;
             }
 
             if (plotType === PlotType.BarChart) {
                 const numContents = receiverX.channel.contents.length;
+                if (numContents > 12) {
```
jorgenherje, line 210: Use const max here?

frontend/src/modules/DistributionPlot/view.tsx
```diff
@@ -252,14 +273,19 @@ export const View = ({ moduleContext, workbenchSettings }: ModuleFCProps<State>)
...
+                setContent(figure.makePlot());
                 return;
             }
 
             if (
                 (plotType === PlotType.Scatter && receiverY.channel) ||
                 (plotType === PlotType.ScatterWithColorMapping && receiverY.channel && receiverColorMapping.channel)
             ) {
+                const numPlots = receiverX.channel.contents.length * receiverY.channel.contents.length;
+                if (numPlots > 12) {
```
jorgenherje, line 285: Use const max here?

frontend/src/modules/SimulationTimeSeriesMatrix/settings.tsx
```diff
@@ -64,6 +66,9 @@ export function Settings({ moduleContext, workbenchSession }: ModuleFCProps<Stat
     const setParameterIdent = moduleContext.useSetStoreValue("parameterIdent");
     const setVectorSpecifications = moduleContext.useSetStoreValue("vectorSpecifications");
 
+    // Transitions
+    const [isPending, startTransition] = React.useTransition();
```
jorgenherje, line 70: Rename to `[isPendingGetParameters, startGetParametersTransistion]` or something similar? To specify what transistion the hook is intendent/utilized for.

jorgenherje (changes requested):
Really nice work! Tried to test most of the features. Code seems very well organized and easy to read!

Just some minor comments from my side.

frontend/src/framework/internal/components/Content/private-components/layout.tsx
```diff
@@ -143,10 +138,26 @@ export const Layout: React.FC<LayoutProps> = (props) => {
...
-        };
+            removeDraggingEventListeners();
+        }
+
+        function calcSizeOfDraggedElement(): Size {
+            const layoutElement = currentLayout.find((element) => element.moduleInstanceId === pointerDownElementId);
+            if (!layoutElement) {
+                return { width: 0, height: 0 };
+            }
+            const rect = convertLayoutRectToRealRect(layoutElement, size);
```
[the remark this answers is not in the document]
rubenthoms: Renamed to `layoutDivSize`.

frontend/src/lib/components/Select/select.tsx
rubenthoms: Good observation, seems like a relict from older days ;). Line 231 would not matter, but I can change it to have it more concise.

frontend/src/lib/utils/geometry.ts
rubenthoms: Renamed a couple of functions, introduced `Vector2D`, and appended `2D` to other types for clarity.

frontend/src/lib/utils/geometry.ts
```diff
@@ -18,51 +23,51 @@ export type Rect = {
...
 }
 
-export function pointDistance(point1: Point, point2: Point): number {
+export function pointDistance(point1: Point2D, point2: Point2D): number {
     return Math.sqrt(Math.pow(point1.x - point2.x, 2) + Math.pow(point1.y - point2.y, 2));
 }
 
-export function pointDifference(point1: Point, point2: Point): Point {
-    return { x: point1.x - point2.x, y: point1.y - point2.y };
+export function vectorDifference(vector1: Vector2D, vector2: Vector2D): Vector2D {
```
jorgenherje, line 38: Change to `pointDifference(point1: Point2D, point2: Point2D): Point2D`, as it is only used with points so far?

frontend/src/lib/utils/geometry.ts
```diff
@@ -18,51 +23,51 @@ export type Rect = {
...
 
-export function pointScale(point: Point, factor: number): Point {
-    return { x: point.x / factor, y: point.y / factor };
+export function vectorScale(vector: Vector2D, factor: number): Vector2D {
+    return { x: vector.x / factor, y: vector.y / factor };
 }
 
-export function pointScaleIndividually(point: Point, factorX: number, factorY: number): Point {
-    return { x: point.x / factorX, y: point.y / factorY };
+export function vectorScaleIndividually(vector: Vector2D, factorX: number, factorY: number): Vector2D {
```
jorgenherje, line 54: Same as above?

frontend/src/lib/utils/geometry.ts
```diff
@@ -18,51 +23,51 @@ export type Rect = {
...
 
-export function pointMultiplyWithScalar(point: Point, scalar: number): Point {
-    return { x: point.x * scalar, y: point.y * scalar };
+export function vectorMultiplyWithScalar(vector: Vector2D, scalar: number): Vector2D {
+    return { x: vector.x * scalar, y: vector.y * scalar };
 }
 
-export function pointMultiplyWithVector(point: Point, vector: Point): Point {
-    return { x: point.x * vector.x, y: point.y * vector.y };
+export function vectorIndividuallyMultiplyComponentsWithOtherVector(vector1: Vector2D, vector2: Vector2D): Vector2D {
```
jorgenherje, line 62: Same as above?

jorgenherje (changes requested):
Perhaps `Vector2D` is not needed when all utility functions are actually called with `Point2D` arguments, and sets states/variables which as of type `Point2D`.

See comments below regarding naming of functions. Seems like most of the `vector...`-functions in `geometry` is not used, and can fine be renamed back to `point...`. 

The original comment regarding `Vector2D` was when an argument was named `vector`, but the type was `Point`. Now it's a bit the other way around, all arguments given to utility functions are `Point2D`, but the function names, arguments and return types are `vector...` and `Vector2D`.

jorgenherje (approved):
Nice! Approved with name fix: `multiply`->`divide` in `geometry.ts`-function as mentioned

Issue #565, opened 2024-01-23
Bugs/issues: Seismic intersection


jorgenherje:
Listing issues/bugs for seismic intersection:

1) MD-label disappears behind the intersection layers
2) When selecting a well and getting "Error loading well trajectories" puts the module in a state where the `esv-intersection`-instance seems to disappear. Re-selecting a well, which was previously shown, show an empty view, with no axes or any `esv-intersection`-info.

-------

**Details**

- Issue 2) above is triggered by following:
     - Select a well an show intersection
 ![image](https://github.com/equinor/webviz/assets/82032112/2eb71f67-b2db-4c38-a372-544c35d458d6)
    - Select an "invalid well"
 ![image](https://github.com/equinor/webviz/assets/82032112/98214ef7-ff91-4268-9e11-f9f3c80009f6)
    - Re-select the first well, and the view does not appear. Data is loaded
 ![image](https://github.com/equinor/webviz/assets/82032112/f2fcf14a-2da0-43f7-b478-d691b54d9121)

rubenthoms:
Closing as this is not relevant anymore.

PR #566, opened 2024-01-24
Only show SMDA wells with survey

HansKallekleiv:
Relates to #565 

=> Only show SMDA wells that have a survey. Remove unused metadata.


jorgenherje (approved):
Looks go to me 👍

PR #576, opened 2024-02-14
Surface improvements

HansKallekleiv:
Various improvements in preparation for more surface stuff, e.g. vds.
- Simplify endpoints and queries
- Speedup resampling when mesh/property surfaces are different
- Add observation surfaces
- Fix camera issues
- Add colors


HansKallekleiv:
Fixed in other pr's.

PR #577, opened 2024-02-14
Refactored `Select` component to fix `Shift` selection bug

rubenthoms:
When releasing shift and pressing it again, the old selection anchor would get lost. The component got refactored and behaves now like the native one. In addition, tests were added for the component and the test framework was adjusted to properly use tailwind.


HansKallekleiv (approved):
Tested and works as expected 👍

PR #579, opened 2024-02-15
Implementation of the GroupTree plugin in the new framework

lindjoha:


anders-kiaer:
Discussed with @lindjoha, we could consider renaming module from "GroupTree" to e.g. "Flow network"

anders-kiaer:
Superseded by another PR 🙂

PR #580, opened 2024-02-15
Implemented framework logic for using `Jotai` atoms

rubenthoms:


anders-kiaer (approved):
LGTM 🤩 Nice to see usage of atoms/`jotai`. Think we will benefit from this both in terms of code structure and performance long term.

jorgenherje (changes requested):
Really nice work! Reviewed as a part of PVT-module refactor. See comments in https://github.com/equinor/webviz/pull/582

jorgenherje (approved):
Looks good to me! 👍

PR #588, opened 2024-02-27
Fix input issues

rubenthoms:
Input was not accepting negative numbers when typed in directly.


HansKallekleiv (approved):
Works perfectly 👍

PR #589, opened 2024-02-29
Revise user sessions

sigurdp:
This PR aims to extend and make our framwork for launching and controlling user sessions more robust. 
It is done in preparation for the upcoming usage of ResInsight as a per-user backend service.

* Added UserSessionManager to help in launching and managing user sessions
* Temporarily added `dev` family of endpoints to allow testing radix related endpoints once logged in.
* Added two user session docker images; `user-mock` as a skeleton/template for user sessions and `user-grid3d-ri` as a placeholder for the upcoming ResInsight user session


backend_py/user_mock/requirements.txt
```diff
@@ -0,0 +1,2 @@
+fastapi
+uvicorn
```
anders-kiaer, line 2: Ideally these (and their sub-dependencies) should be pinned using e.g. `poetry` like we do for the primary backend, so we can rebuild without surprise.

Fix it now or later?

backend/src/services/user_session_manager/user_session_manager.py
```diff
@@ -0,0 +1,309 @@
...
+    num_calls = 1
+
+    session_info = session_dir.get_session_info(job_component_name, instance_str)
+    if not session_info:
+        return None
+
+    # Given that we found info on the session and it's not in the running state, we will try and wait for it to come online.
+    # The job/session might be in the process of being created and spinning up, so we will try and wait for it.
+    # We will need to add a timeout to this!!!!!
+    # How much time should we spend here before giving up?
```
anders-kiaer, line 128: Is startup time of the docker image neglible here, if so it is a question we could ask Radix? Maybe they have a timeout already before a session/job is marked as failed(-to-provision)?

anders-kiaer (approved):
Impressive work. A couple of comments (dependency management being the most important question), otherwise LGTM 👍

backend_py/user_mock/requirements.txt
sigurdp: Indeed they should, and for the ResInsight/user_grid3d_ri component we'll certainly employ Poetry once we get going.
As for the user_mock component, I suggest we revisit it when the structure of `user_grid3d_ri` has been settled. 
In the meantime I'll pin the versions of `fastapi` and `uvicorn` in the `requirements.txt` file for the mock component.

backend/src/services/user_session_manager/user_session_manager.py
sigurdp: Removed the stale comment and added som more detailed comments on timeouts while we seek the truth either through practical usage or from the Radix team.

PR #595, opened 2024-03-14
Skeleton for grid3d frontend and backend

sigurdp:
* Disabled the Grid3D frontend modules until we have some sensible content
* Renamed primary router folder/prefix from grid to grid3d
* Added new Grid3DVTK frontend module as a copy of existing Grid3D module
* Added skeleton job component for user-grid3d-vtk
* Use poetry in both user_grid3d_vtk and user_mock


HansKallekleiv (approved):
👍

PR #596, opened 2024-03-19
Support 3D grids using ResInsight

sigurdp:
Prototype implementation of support for 3D grids, leveraging ResInsight in a user session as the computational backend


HansKallekleiv (approved):
🥇

PR #618, opened 2024-05-02
Use elastic search directly for timeseries column names

HansKallekleiv:
For discussion.

- Use elastic search directly instead of the sumo explorer
- Get the column names from the first valid realization instead of the collection.

Reduces the time spent from 15-20s to 0.5s.
This is probably due to how how the explorer/sumo wrapper uses batching when handling large document collections from elastic search. Should be investigated.


anders-kiaer:
Superseded by another PR

PR #619, opened 2024-05-06
Fix bug in RealizationPicker component

jorgenherje:
Fix bug for realizaiton picker, allow realization number 0 in input. Realization number = 0 gave error for both single value and range `a-b` input.


rubenthoms (approved):
LGTM

PR #629, opened 2024-06-04
inplace quickfix

HansKallekleiv:


HansKallekleiv:
Done in https://github.com/equinor/webviz/pull/630

PR #630, opened 2024-06-04
Refactoring of Sumo access classes to prepare for more direct usage of elastic search

sigurdp:
Refactored the entire set of Sumo access classes in preparation for more direct usage of elastic search.

This means that we will try to untangle a bit from the Sumo Explorer, and will instead try and utilize direct searches against Elasticsearch as well as alternative download methods where it makes sense from a performance standpoint. 

This PR mainly rearranges our Sumo access classes to cater for this hybrid usage, but also introduces a specific fix the slow fetching of available summary vector names. This is now implemented using a direct Elasticsearch query which gives significant speedups.


HansKallekleiv (approved):
👍

PR #636, opened 2024-06-10
Work around bug in Sumo Explorer that reports too few iterations/ensembles

sigurdp:
There is a bug in Sumo Explorer's `Case.iterations_async`
Revert back to using the sync version of `Case.iterations` instead of `Case.iterations_async` until this has been remedied.

See also: https://github.com/equinor/fmu-sumo/issues/326


anders-kiaer (approved):
👍 LGTM!

PR #642, opened 2024-06-13
Fix first time user challenges

### LINKED ISSUE #626: Persist selected field in Ensemble selector dialog box
It is common that a user has access to several fields, but for a given webviz session he/she are often only working on a given field.

In the ensemble selector, the dropdown list of selected fields resets to first field in list every time the ensemble dialog box opens up when user is adding/removing ensembles. 

One suggestion is to e.g. persist last selected field, and use that as default in the selected field dropdown?

Need to handle the case that the earlier persisted field might no longer be a valid selection in the dropdown (changes in SUMO and/or SUMO access), and e.g. fall back to first field as today in those cases.

### LINKED ISSUE #637: First-time user UX challenges
Observing first-time users have given this feedback:

1. When entering the application, some do not immediately understand that in order to use modules they need to select some ensembles (instead some users add modules, like Simulation Time Series, which then shows no data and they struggle with understanding where to click further in order to add actual data to the module).
image

2. When adding a new module, some do not immediately understand they need to click on the added module in order to go further and populate settings in order to show data.

### LINKED ISSUE #644: Narrow (few pixels) are for scroll bar vs. adjusting settings width
Some users apparently prefer scrolling settings tab with scroll bar (when height is bigger than window).

Currently, it is only a couple of pixels where they can use the scroll bar (moving 1-2 pixels too far to the right changes the cursor into adjusting settings width mode).

image

rubenthoms:
Closes #626.
Closes #637.
Closes #644.


anders-kiaer (approved):
Tested, LGTM! As discussed, the green message box is a temporary solution until we know how to best introduce/open the settings tab.

PR #651, opened 2024-06-17
Intersection fixes

HansKallekleiv:
- Fixed Seismiclayer bounding box not being recalculated when switching polyline
- Removed unused files/functions


rubenthoms (approved):
:+1:

PR #660, opened 2024-06-27
Parameter fixes

HansKallekleiv:
- Set max plots to 56
- Filter out log10 in module for now. revisit when metadata is matured.


anders-kiaer (commented):
Should we keep the `LOG10` information as information regarding which axis to use? I.e. if parameter X has a sibling `LOG10-X` then we use logarithmic scale for `X` when plotting with plotly?

I guess later this information will be derived easily from metadata on parameters wrt. which distribution type it comes from.

anders-kiaer (approved):
👍 LGTM

anders-kiaer (approved):
👍🏻

PR #667, opened 2024-07-15
Refactored and extended existing client side 2d vector functionality

sigurdp:
* Split 2d vector functionality out into separate file (from existing `geometry.ts`)
* Added functions for rotating a vector and a point
* Refactored/renamed some of the functions


frontend/src/lib/utils/vector2.ts
```diff
@@ -0,0 +1,71 @@
...
+export function addScalarToVector(vector: Vector2, scalar: number): Vector2 {
+    return { x: vector.x + scalar, y: vector.y + scalar };
+}
+
+export function vectorsEqual(vector1: Vector2, vector2: Vector2): boolean {
+    return vector1.x === vector2.x && vector1.y === vector2.y;
+}
+
+// Rotate vector counter-clockwise by the given angle (in radians)
+export function rotateVector(vec: Vector2, angle: number): Vector2 {
```
rubenthoms, line 56: Consider calling `angle` `angleInRad` (CCW should be assumed as it's the default) to avoid any misunderstandings. Same below.

rubenthoms (approved):
One comment, otherwise good to go.

frontend/src/lib/utils/vector2.ts
sigurdp: Renamed parameter

PR #668, opened 2024-07-16
Add utility functions for encoding/decoding list of uints and key-value pairs in URL-friendly strings

sigurdp:
Added utility functions for encoding list of uints and key-value pairs in URL friendly strings


rubenthoms (approved):
Looks good to me :+1: 

Ready to merge after committing the few points we discussed.
- null
- Record<...>
- throw on double `~`
- throw on unsupported type in property bag encoding
- `unsignedIts` -> `unsignedInts`

PR #671, opened 2024-07-29
Implemented persistence of switch values in ensemble dialog

### LINKED ISSUE #500: Persist selected toogles in Ensemble selector
We currently have this as default:
image

This is a good default for users going into Webviz, who intends to look at some of the more official simulation results (e.g. management).

For modelling team members however, a different combination is more useful where they can see all cases (and maybe only the ones they have created):
image

Today they have the freedom of choosing what they want, but their choice is not persisted when closing/opening the Ensemble dialog box. 

---

A good solution here might be:
- Let the current toggle persist when closing/opening the ensemble dialog box (i.e. current toggle setting persist throughout the session).
- When you open a new browser tab (i.e. start a new session), the default value would be the last value actively set by the user in a previous session, and stored in `localStorage`.

This pattern is similar to #499.

rubenthoms:
Closes #500.

### DIFF
--- a/frontend/src/framework/internal/components/SelectEnsemblesDialog/selectEnsemblesDialog.tsx
+++ b/frontend/src/framework/internal/components/SelectEnsemblesDialog/selectEnsemblesDialog.tsx
@@ -50,7 +50,7 @@ interface CaseFilterSettings {
     users: string[];
 }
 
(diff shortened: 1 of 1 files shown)


jorgenherje (approved):
👍

PR #673, opened 2024-07-29
Improve ensemble color pre-selection

rubenthoms:
Initial color assignment in ensemble dialog is now taking into account colors that are already in use.


jorgenherje (approved):
👍

PR #682, opened 2024-08-02
Remove/ignore SMDA wells that has undefined entries in trajectory

HansKallekleiv:
Remove SMDA wells that has undefined (nan) entries in the trajectory.


rubenthoms (commented):
The two functions look quite similar to a great extent. Should we make a shared common function?

HansKallekleiv:
> The two functions look quite similar to a great extent. Should we make a shared common function?

Fixed. 

Also removed an unused argument and fixed wells being upside down in SubsurfaceMap. This module will be removed though, when we have finished 2dviewer and 3dviewer modules.

PR #685, opened 2024-08-13
Implementation of `SortableList` component (+ sub components) and menu utility components

rubenthoms:
Added new components to `lib` in order to use them in multiple modules such as `Intersection`, `3D viewer`, `2D viewer`, and `Well log viewer`. The main component is a `SortableList` which can have children of type both `SortableListItem` and `SortableListGroup`. The latter can also wrap other `SortableListItem` and `SortableListGroup` components. Whenever the order of the list is changed, the `onItemMoved` callback is called with its four arguments:

```typescript
itemId: string // the ID of the item which was moved
originId: string | null // the ID of the current parent of the item which was moved - null when main list (top level), string when a SortableListGroup component
destinationId: string | null // the ID of the new parent after the move operation is performed - null when main list (top level), string when a SortableListGroup component
position: number // (0 - number of parent's children + 1) - position in the destination component where the moved item was positioned
```

The groups and items of the list can be individually filled with adornments, e.g. action buttons (start: before title - end: after the title).


frontend/src/lib/components/SortableList/sortableList.tsx
```diff
@@ -0,0 +1,645 @@
...
+                    onItemMoved(draggedElement.id, originId, destinationId, position);
+                    return;
+                }
+
+                const originId = getItemParentGroupId(draggedElement.element);
+                const destinationId = getItemParentGroupId(currentlyHoveredElement.element);
+                const positionDelta = currentlyHoveredElement.area === HoveredArea.TOP ? 0 : 1;
+                const position = getItemPositionInGroup(currentlyHoveredElement.element) + positionDelta;
+
+                onItemMoved(draggedElement.id, originId, destinationId, position);
```
Anders2303, line 527: If you move an item to the spots above or below itself, the item order wont actually change, but the `onItemMoved` event will still fire.

Wrote a simple check [for that in the welllog-viewer's code](https://github.com/equinor/webviz/blob/well-log-viewer/frontend/src/modules/WellLogViewer/settings/components/CurveTracks/TrackSettings.tsx#L122..L124), but that should prefferably be moved into this comp instead

```typescript
const currentPosition = props.plots.findIndex((p) => p.name === movedItemId);
if (currentPosition === newPosition || currentPosition + 1 === newPosition) return;
```

frontend/src/lib/components/SortableList/sortableListItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { HoveredArea, SortableListContext } from "./sortableList";
+import { SortableListDropIndicator } from "./sortableListDropIndicator";
+
+export type SortableListItemProps = {
+    id: string;
+    title: React.ReactNode;
+    initiallyExpanded?: boolean;
+    startAdornment?: React.ReactNode;
+    endAdornment?: React.ReactNode;
+    children: React.ReactNode;
```
Anders2303, line 17: In the well-log viewer, the sortable items for the individual plots dont have any child content; the children prop should be optional for items that don't intend to expand

frontend/src/lib/components/SortableList/sortableListItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+    const sortableListContext = React.useContext(SortableListContext);
+
+    const isHovered = sortableListContext.hoveredElementId === props.id;
+    const isDragging = sortableListContext.draggedElementId === props.id;
+    const dragPosition = sortableListContext.dragPosition;
+
+    function handleToggleExpanded() {
+        setIsExpanded(!isExpanded);
+    }
```
Anders2303, line 45: This is never called, since the generic toggle-adornment was removed from `<Header>`. Should either make "expanded" be a prop, leaving it fully up to the implementer; or make it possible for adornments to use this callback (and probably having a default toggle-adornment if so)

Anders2303 (commented):
Noted down some smaller things I ran into when implementing this in the Well Log Viewer branch

Anders2303 (approved):
Looks good to me 👍

PR #700, opened 2024-08-21
Fix multi select initial anchor issue

### LINKED ISSUE #699: Select button not working initially in multi select list
How to reproduce:

- Select three ensembles in ensemble dialog box
- Add a new instance of `SimulationTimeSeries`
- Try to select all three ensembles by keeping select button down, observe this does not work.
- In order to select all, you can keep ctrl down while selecting them.
- ...but after using ctrl: suddenly using select button during selection works as expected again.

rubenthoms:
Closes #699.


anders-kiaer (approved):
👍

PR #710, opened 2024-09-03
Endpoint tests with Sumo data

### LINKED ISSUE #548: Integration tests setup
### Details
- Currently, we are not able to write/execute integration tests due to lacking implementation of automatic authentication for the test user
- SUMO offered us to help us with this (https://github.com/equinor/fmu-sumo/issues/257)
- There are some best practices described in https://github.com/equinor/appsec-fundamentals-authn-authz/blob/main/ex-14/doc/integration_testing.md
- Already described in #485 

- Anders asks Per Olav about how/if they implemented integration tests and if we can use the same pattern

HansKallekleiv:
Closes #548 

- Added new github actions workflow for testing against Sumo prod with shared key
- Added some endpoint tests for timeseries and explore
- Rewrote explore router to be similar as other routers
- Bump pytest and added pytest-timeout and pytest-asyncio plugins
- Removed non-working parameteraccess tests


.github/workflows/webviz.yml
```diff
@@ -106,7 +106,7 @@ jobs:
           black --check primary/ tests/
           pylint primary/ tests/
           bandit --recursive primary/
-          mypy primary/ tests/
+          mypy primary/
```
jorgenherje, line 109: Why is test-folder excluded now?

jorgenherje (commented):
Was there any conclusion on the issue of regarding PRs has to come from the equinor/webviz repository in order to run the endpoint test actions?

.github/workflows/backend_sumo_prod.yml
```diff
@@ -0,0 +1,63 @@
...
+  push:
+  pull_request:
+    branches:
+      - main
+  release:
+    types:
+      - published
+  # workflow_dispatch:
+  # schedule:
+  #   - cron: "48 4 * * *"
```
anders-kiaer, line 13:
```suggestion
  push:
  workflow_dispatch:
  schedule:
    - cron: "0 0 * * *"
```
Due to the secrets not working for incoming PRs (from forks), we for now limit to only runs initiated from the repo itself (push to branch in official repo, workflow dispatch and schedule).

anders-kiaer (approved):

PR #726, opened 2024-09-17
Bugfix: `Intersection` `Wellpicks` layer re-render issue

### LINKED ISSUE #724: Slowdown in `Intersection` module's Well-pick layer
Adding a well-pick layer to the `Intersection` module causes significant frame-drops

* If the module's settings are open, the settings component's Tree RC count keeps increasing, so there's a re-render loop being triggered somewhere
* Additionally, just having the well-pick layer enabled, cause the entire app to behave slower (without Intersection's settings being open). The view's debug info is stable, so the slowdown is happening somewhere else, but you can clearly see things slowing down when moving the graph around

rubenthoms:
Closes #724.

Also fixes issue preventing data fetching when layer is added.


Anders2303 (approved):
Looks good to me. 👍

Issue #722, opened 2024-09-12
Improve Realization Filtering


jorgenherje:
Improve functionality of realization filtering. Enhanced user interaction. 

Provide two ways of selecting subset of realizations:
- Select realization numbers
- Select parameter and value range -> create subset of realization numbers matching the range
    - Multiple parameters -> &&-logic
- Include/exclude functionality
- Keep filtering by selecting realization numbers
- Add visualization of available and selected realization numbers
    - Interactive/Graphical selection for realization numbers
    - Visualize non-existing realization numbers as grey/transparent with dashed border, to show these are non-existing. Available realizations can be e.g. grey, faded blue. Selected realizations are green/blue?
- Add filtering based on parameter values (slider for parameter value range)
   - Visualizing/preview with selected parameter and value range

----
- Visualization of selected realization numbers can be inspired by sumo.
- 

Related issues:
- #690 
- #691 
- #321

Issue #691, opened 2024-08-19
Realization filter: Rename from `index` to `number` in naming


jorgenherje:
The realizations are known as number, not index. Rename and remove naming `index`, as it can be misguiding.

![image](https://github.com/user-attachments/assets/c6ff35fa-bc24-4190-9792-f607ae64cfc3)

---

Also add placeholder text for dropdown if possible:
![image](https://github.com/user-attachments/assets/f63d14bf-5a8f-4166-ab83-065c1bc00db3)

Issue #690, opened 2024-08-16
Inform user about available realization set to choose from in realization filter


anders-kiaer:
![image](https://github.com/user-attachments/assets/5288b798-e075-40f5-9dbe-81c043ffaae3)

Maybe we can do that by also offering a "rendering" of the currently selected set of realizations, with colored circles (to be consistent with SUMO). Then the user also have the option of either giving text input, or selecting the subset of realization graphically:

![image](https://github.com/user-attachments/assets/c586b8ae-06ed-4d91-8bd9-88dee4881343)

Issue #321, opened 2023-09-20
Provide functionality for parameter filtering


jorgenherje:
When using ensemble parameters or a union of parameters across ensembles, the list can become large. Consider creating component/class/hook/utility functionality to be able to set some filtering criteria and provide a list of parameters fulfilling the filter selection for usage.

E.g. component to set based on wanted `EnsembleParameter` settings/attributes - see: https://github.com/equinor/webviz/blob/main/frontend/src/framework/EnsembleParameters.ts

Types:
* Filter of `Parameter` to create shorter list of parameter names to select
* Filter of `Parameter` to create a subset of realizations to look at

-----

Mentions:

* Hans' [comment](https://github.com/equinor/webviz/pull/320#pullrequestreview-1638447023) 
   * The parameter list will be overwhelming if you don't know what to look for. Not directly related to this PR but we need both filtering of parameters(e.g. based on group) and sorting of parameters on how "important" they are. We can do this by calculating the correlation of the parameter values to the timeseries at a given date. I think we could automate this by selecting the last date (automatically) and perform the correlation. There exists an early attempt of an endpoint on the api to do this correlation: https://github.com/equinor/webviz/blob/main/backend/src/backend/primary/routers/correlations/router.py
   
* My [comment](https://github.com/equinor/webviz/pull/320#issuecomment-1730912897):
   * Three types of parameter filtering?: 
      - Filtering based on criteria/attributes of [`Parameter`](https://github.com/equinor/webviz/blob/main/frontend/src/framework/EnsembleParameters.ts) type to provide a shorter list of  parameter names in e.g. a dropdown.
      - Filtering based on values of a `Parameter`, i.e. functionality as mentioned above here - providing a shorter list.
      - Filtering based on criteria which results in a subset of wanted realization numbers?

HansKallekleiv:
Sounds like a good list.
Some comments on the last two:

>     * Filtering based on values of a `Parameter`, i.e. functionality as mentioned above here - providing a shorter list.
This is more a ranking/soring of the parameters based on how "important" or correlated they are.
It can also be used as filtering by removing parameters with little correlation.
This can also be visualized as done in the existing ParameterAnalysis plugin in webviz-subsurface.

![image](https://github.com/equinor/webviz/assets/16436291/c6370dbf-392d-40fe-b5c1-5cdd96610932)


>     * Filtering based on criteria which results in a subset of wanted realization numbers?
Yes. Filterting based on parameter values => Subset of realizations => Updated correlations
[...]

PR #741, opened 2024-10-02
Upgrade realization filtering

jorgenherje:
This PR upgrades the RealizationFilter, by including a filter for each selected ensemble.

**Upgrades to Realization Filter:**
- Realization filter drawer with one ensemble realization filter component for each selected ensemble
- Can filter by parameter value, i.e. select continuous or discrete parameters (continuous and discrete), and select value range/discrete values to get realizations which matches these selections (intersection of all matches), i.e. &&-logic.
   - Slider to select range for continuous parameter values
   - Picker to select set of discrete values to match
- Visualization of which realizations will be selected based on the current filter selections - display component.

**SmartNodeSelector:**
- Provide valid state with the selected tags for `onChange` data
- Fix selected `nodeIds` to be correct when using wildcard
- Prevent `onChange` to trigger when clicking outside of components

----

The PR closes the following issues:

Closes: #690 
Closes: #691
Closes: #321 
Closes: #722


HansKallekleiv (approved):

anders-kiaer (approved):
Nice work. One small suggestion given offline regarding if we should show tooltip also in the default smaller rendered ensemble circles. Overall LGTM. 👍

frontend/src/framework/utils/realizationFilterTypesUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+ * DiscreteParameterValueSelection is a union type of string[] and number[].
+ */
+export function isArrayOfNumbers(discreteValues: DiscreteParameterValueSelection): discreteValues is readonly number[] {
+    if (discreteValues.length === 0) {
+        return true;
+    }
+
+    // Check first element only for efficiency, as input is string[] | number[]
+    return typeof discreteValues[0] === "number";
+}
```
Anders2303, line 59: I think these functions are fairly general and potentially useful in other module code. Should we move it to a global `framework/utilis` file?

frontend/src/framework/utils/realizationFilterTypesUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+): number[] {
+    if (!selections) return [];
+
+    const realizationNumbers: number[] = [];
+    for (const selection of selections) {
+        if (typeof selection === "number") {
+            realizationNumbers.push(selection);
+        } else {
+            realizationNumbers.push(
+                ...Array.from({ length: selection.end - selection.start + 1 }, (_, i) => selection.start + i)
```
Anders2303, line 122: Could alternatively use the lodash utility [`range()`](https://lodash.com/docs/4.17.15#range) for this, which makes it a bit more obvious what you're doing

```js
_.range(selection.start + 1, selection.end)
```

frontend/src/lib/components/SmartNodeSelector/smartNodeSelector.tsx
```diff
@@ -952,29 +960,34 @@ export class SmartNodeSelectorComponent extends React.Component<SmartNodeSelecto
...
+                    selectedIds.push(matchedIds[j][matchedIds[j].length - 1] ?? "");
                 }
             }
         }
         if (
             !this.selectedNodes ||
             selectedNodes.length !== this.selectedNodes.length ||
-            JSON.stringify(this.selectedNodes) !== JSON.stringify(selectedNodes)
+            JSON.stringify(this.selectedNodes) !== JSON.stringify(selectedNodes) ||
+            JSON.stringify(this.props.selectedTags) !== JSON.stringify(selectedTags)
```
Anders2303, line 990: Optionally, could use lodash's [`isEqual()`](https://lodash.com/docs/4.17.15#isEqual) here

jorgenherje: Perhaps true. What do you think @rubenthoms ?

rubenthoms: `SmartNodeSelector` comes from a repository (wcc) where we did not have `lodash` as a dependency. That's why it hasn't been used in the component so far. Since we have our own implementation copy of it in the Webviz NG repository (due to Tailwind styling), we could of course start replacing all these occurrences with `isEqual` as it is generally more performant. However, since we might (although with low probability) apply the fixes from Webviz in wcc, it seems beneficial to have both implementations use the same code for now. Either we stick to the no-lodash variant until all ties to wcc are really cut or we refactor both versions to using `isEqual`. This, however, does not seem to be one of the important tasks at the moment as the lower performance has not caused any issues so far.

jorgenherje: We'll leave it as it is for now :)

Issue #743, opened 2024-10-07
Fix spacing/gap in RadioGroup component


jorgenherje:
The spacing/gap between radio items in a radio group is a bit too large. Takes up too much space.

![image](https://github.com/user-attachments/assets/a2065df0-b008-4eb0-b9ac-d9745e548712)

PR #744, opened 2024-10-07
Adjust gap for RadioGroup component

jorgenherje:
Decrease the gap, as items takes up too much space.

Vertically: gap-1 seems to be a reasonable gap
Horizontally: gap-3 seems suitable to separate text of a radio group element to the radio button of next radio group element.

Closes: #743


Anders2303 (approved):
Looks good to me

Issue #754, opened 2024-10-11
Add secondary color to iconButton


jorgenherje:
Add secondary color for icon button, as done for regular button component

Issue #753, opened 2024-10-11
Improve UX for Color Realization By Parameter


jorgenherje:
Coloring by parameter is only available when visualizing Individual Realizations, but it is not very easy to see.

Improve UX s.t. it is more clear for user and interaction becomes more intuitive.

PR #755, opened 2024-10-11
Improve UX for Color realization by parameter in SimulationTimeSeries

jorgenherje:
Improve UX for coloring realizations by parameter value in SimulationTimeSeries-module.

Make dependency between parameter selection and active Individual realization more intuitive.

Also add secondary color for iconButton

---
Now placed together with Visualization selections radio buttons 

image

----

Closes: #753
Closes: #754


anders-kiaer (approved):
LGTM 👍

Issue #683, opened 2024-08-07
Add `WellLogViewer` module


rubenthoms:
Create a new module named `WellLogViewer` ("Well log viewer"). Use the `@webviz/well-log-viewer` component located in the `webviz-subsurface-components` repository (https://github.com/equinor/webviz-subsurface-components/tree/master/typescript/packages/well-log-viewer).

Specifications (1st iteration):
- [x] Settings
  - [x] Dropdown for selecting ensemble
  - [x] Load available wellbores for field given by selected ensemble from backend
  - [x] Dropdown for selecting wellbore
  - [x] Definition of logs (columns) - `Intersection` module with its layer manager could give some inspiration
- [ ] View
  - [ ] Use `well-log-viewer` component to visualize well log defined in settings
  - [ ] Load data required for visualization from backend
- [ ] Define interface for sending log definitions, ensemble, and wellbore from settings to view

Specifications (2nd iteration):
- [ ] Settings
  - [ ] Make it possible to sync selected wellbore and ensemble with other modules
- [ ] View
  - [ ] Sync hovered MD with and subscribe to MD hover from other modules

Specifications (3rd iteration):
- [ ] Settings
  - [ ] Make definitions of log storable in and selectable from local storage (maybe in addition or alternatively savable as/loadable from JSON file)

Specifications (4th iteration):
- [ ] Evaluate if component created by Aspentech is sufficient - maybe it requires refactoring to fit into our application (e.g. remove legend and primary scale and make it possible to define these on the outside via props/callbacks)

PR #707, opened 2024-08-30
Replace Intersection and 3D viewer readout boxes with shared comp

Anders2303:
* Moved the `EsvIntersection` and `VectorSelect` components from `framework/components` to `modules/_shared`
* Replaced the ReadoutBox components in `EsvIntersection` and `3dViewer` with a shared `ReadoutBoxc` component
  * Also added logic to have the readout-box move away from the mouse by flipping it's position.

PR #702, opened 2024-08-22
Implement Well Log Viewer Module

Anders2303:
Resolves issue #683 

Adds a wrapper module for the [Subsurface WellLogViewer](https://github.com/equinor/webviz-subsurface-components/tree/master/typescript/packages/well-log-viewer).

### TODOs
This uses the readout box and sortable list components introduced in other PRs, those should be reviewed and merged first
- [x] Await merge of #685 
- [x] Await merge of  #707 
- [x] Await merge of #717 

Specifications (1st iteration):
- [x] Settings
  - [x] Dropdown for selecting ensemble
  - [x] Load available wellbores for field given by selected ensemble from backend
  - [x] Dropdown for selecting wellbore
  - [x] Definition of logs (columns) - `Intersection` module with its layer manager could give some inspiration
- [x] View
  - [x] Use `well-log-viewer` component to visualize well log defined in settings
  - [x] Load data required for visualization from backend
- [x] Define interface for sending log definitions, ensemble, and wellbore from settings to view

Specifications (2nd iteration):
- [x] Setting sync
  - [x] ~~Ensemble~~
     - _Skipped this one, we dont have any existing setup for synced field_ 
  - [x] Wellbore
- [x] View
  - [x] Show curve info on hover
    - Subsurface component is currently not emitting data properly, see https://github.com/equinor/webviz-subsurface-components/issues/2196 . Will fix that first, before implementing this
  - [x] Sync hovered MD with and subscribe to MD hover from other modules
- [ ] Colors
  - [x] Avoid random colors each re-render
  - [ ] Make colors update when workbench color settings are changed
  - [ ] User color selection 

Specifications (3rd iteration):
- [ ] Extra curve types
  - [x] Diff curves. 
  - [ ] Discrete curves. The viewer supports it, but requires a "discrete_metadata" field for the well-log. How do we fetch or set up these?
- [ ] Settings
  - [ ] Map out other settings in the subsurface component, and see what makes sense to expose for our users
    - [x] Horizontal/vertical orientation
    - [ ] Curve-specific scales
  - [x] Make definitions of log storable in and selectable from local storage (maybe in addition or alternatively savable as/loadable from JSON file)

Specifications (4th iteration):
- [ ] Evaluate if component created by Aspentech is sufficient - maybe it requires refactoring to fit into our application (e.g. remove legend and primary scale and make it possible to define these on the outside via props/callbacks)

PR #794, opened 2024-11-04
Implement Well-log viewer module MVP

Anders2303:
_This PR is equivalent to #702; this is copy-branch to resolve some stuff that broke during some rebases_

### Branch dependencies
This uses the readout box and sortable list components introduced in other PRs, those should be reviewed and merged first
- [x] Await merge of #685 
- [x] Await merge of  #707 
- [x] Await merge of #717 

(Partially) resolves issue #683 

Adds a wrapper module for the [Subsurface WellLogViewer component](https://github.com/equinor/webviz-subsurface-components/tree/master/typescript/packages/well-log-viewer). The component renders out well-log data along a given wellbore's MD/TVD values. Note that this implementation is a simpler MVP, an is therefore not feature-complete; more features will be introduced in separate PRs.

Features:
- Fetch data from `well/log_curve_data` and `well/wellbore_log_curve_headers` API endpoint
  - (Aka, the SSDL wellbore-curve APIs)
- Show _continuous_ data plots
  - Line, area, gradient, dot, and differental plots are supported
- Show vertical lines for Well-picks
- Synced settings:
  - Global Hover MD
  - Wellbore
  - Vertical Scaling
- Export/import JSON file of track config
- Persistent setting across reloads

### Future Work
- Add support for discrete (or "stacked") plots
  - Implemented in a local branch, PR to come<LINK TO DRAFT PR>
- More graph data sources:
  - Geology and Stratigraphy on SMDA
     - Implemented in local branch, PR to come <LINK TO DRAFT PR>
  - HAZI, DEVI, DSL curves
     - See SSDL "wellbore survey" for these  
- More settings for visuals
  - Plot domain (visual min-max)
  - Color settings
  - Stacked graph settings


frontend/src/modules/WellLogViewer/view/SubsurfaceLogViewerWrapper.tsx
```diff
@@ -0,0 +1,265 @@
...
+    // Viewer config
+    horizontal: boolean;
+    padDataWithEmptyRows: boolean;
+    templateTracks: TemplateTrack[];
+
+    // Passing the module props to make context and service access less cumbersome
+    moduleProps: ModuleViewProps<InterfaceTypes>;
+};
+
+function useGloballySyncedMd(
```
rubenthoms, line 56: Misleading name, isn't it?

frontend/src/modules/WellLogViewer/view/SubsurfaceLogViewerWrapper.tsx
```diff
@@ -0,0 +1,265 @@
...
+            const payload: GlobalHoverMd = newMd ? { wellboreUuid, md: newMd } : null;
+
+            workbenchServices.publishGlobalData("global.hoverMd", payload, instanceId);
+        },
+        [instanceId, wellboreUuid, workbenchServices]
+    );
+
+    React.useEffect(() => {
+        return workbenchServices.subscribe("global.hoverMd", handleGlobalValueChange, instanceId);
+    }, [handleGlobalValueChange, instanceId, workbenchServices]);
```
rubenthoms, line 91: Why not handling everything inside the useEffect instead of creating a callback?

frontend/src/modules/WellLogViewer/view/SubsurfaceLogViewerWrapper.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+    return broadcastGlobalMdChange;
+}
+
+function useGloballySyncedVerticalScale(
```
rubenthoms, line 96: Same here about the name.

frontend/src/modules/WellLogViewer/view/SubsurfaceLogViewerWrapper.tsx
```diff
@@ -0,0 +1,265 @@
...
+        if (verticalSyncActive) {
+            const unsubscribe = workbenchServices.subscribe(
+                "global.syncValue.verticalScale",
+                handleGlobalVertScaleChange,
+                moduleInstanceId
+            );
+
+            return unsubscribe;
+        }
+    }, [workbenchServices, verticalSyncActive, syncableSettingKeys, moduleInstanceId, handleGlobalVertScaleChange]);
```
rubenthoms, line 127: Same as above.

frontend/src/modules/WellLogViewer/view/SubsurfaceLogViewerWrapper.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+    const handleTrackMouseEvent = React.useCallback(
+        (/* welllogView: WellLogView, e: TrackMouseEvent */) => {
+            // ! No-op method. Passed to the viewer to make it not show the context menu for tracks
+        },
+        []
+    );
+
+    return (
+        // The weird tailwind-class hides the built-in hover tooltip
```
rubenthoms, line 237: Any chance to implement that as a feature in the `WellLogViewer`?

frontend/src/modules/WellLogViewer/view/ReadoutWrapper.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { TemplateTrack } from "@webviz/well-log-viewer/dist/components/WellLogTemplateTypes";
+
+import _ from "lodash";
+
+import { DEFAULT_MAX_VISIBLE_TRACKS } from "../utils/logViewerTemplate";
+
+export type ReadoutWrapperProps = {
+    templateTracks: TemplateTrack[];
+    wellLogReadout: Info[];
+    hide?: boolean;
```
rubenthoms, line 14: Is hide ever used? You seem to already hide it in `SubsurfaceLogViewerWrapper` by not rendering it.

rubenthoms (changes requested):
Nice work! :+1:

Admittedly, I have quite a few comments and suggestions but this is mostly due to the PR having gotten quite large.

backend_py/primary/primary/services/ssdl_access/types.py
```diff
@@ -34,18 +35,33 @@ class WellborePerforation(BaseModel):
...
-    curve_alias: str
-    curve_description: str
+    curve_alias: str | None
+    curve_description: str | None
     index_unit: str
-    no_data_value: float
+    no_data_value: float | None
+    unit: str
+    curve_unit_desc: str | None
+    DataPoints: list[list[float | None]]
```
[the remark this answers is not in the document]
Anders2303: That casing is just how SSDL has decided to structure the JSON you get from them.
I think the casing for anything in this type file **should** match the expected payloads 1:1, as it makes writing the model-validations more straight-forward (`WellboreLogCurveData.model_validate(ssdl_data)`). I'll add a comment to explain why it's in the wrong case

Optionally, could _also_ add a property that let's people refer to the data with the correct casing after validation

```py
class WellboreLogCurveData(BaseModel):
    # .... other props
    # This field has weird casing. This is just how SSDL has decided to return this object, so we leave it as is
    DataPoints: list[list[float | None]]
    @property
    def data_points(self) -> list[list[float | None]]:
        # Utility property to let us refference "DataPoint" with proper casing
        return self.DataPoints

```

frontend/src/lib/components/Dropdown/dropdown.tsx
```diff
@@ -20,6 +20,7 @@ export type DropdownOption<TValue = string> = {
     adornment?: React.ReactNode;
     hoverText?: string;
     disabled?: boolean;
+    group?: string;
```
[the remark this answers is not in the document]
Anders2303: Issue created, see #816

frontend/src/modules/WellLogViewer/settings/atoms/derivedAtoms.ts
```diff
@@ -0,0 +1,107 @@
...
+    userSelectedWellboreUuidAtom,
+} from "./baseAtoms";
+import { logViewerTrackConfigs } from "./persistedAtoms";
+import {
+    drilledWellboreHeadersQueryAtom,
+    wellLogCurveHeadersQueryAtom,
+    wellborePicksAndStratigraphyQueryAtom,
+} from "./queryAtoms";
+
+export const selectedEnsembleSetAtom = atom((get) => {
```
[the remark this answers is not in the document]
Anders2303: Filtered doesn't seem quite right either. What's wrong is the use of "set" when I'm actually picking one from the set, so I'll rename it to `selectedEnsembleAtom`

frontend/src/modules/WellLogViewer/settings/components/AddItemButton.tsx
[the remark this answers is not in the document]
Anders2303: I'll make a note for that on the v2 PR

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/SortablePlotList.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+    // If the current selection does not exist, keep it in the selection, with a warning. This can happen when the user is importing a config, or swapping between wellbores
+    allSelectedWellLogCurves.forEach((curveName) => {
+        if (!curveHeaderOptions.some(({ value }) => value.endsWith("::" + curveName))) {
+            curveHeaderOptions.push(makeMissingCurveOption(curveName));
+        }
+    });
+
+    // TODO, do an offsett or something, so they dont always start on the same color?
+    const colorSet = React.useRef<ColorSet>(new ColorSet(CURVE_COLOR_PALETTE));
```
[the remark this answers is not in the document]
Anders2303: I'll leave color selection for a later pr.... I think I'll need some bigger rewiring to make it properly synced with the workbench settings, since color, color2, fill, fill2, and colorTable all need to be set differently for different curve-types

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/SortablePlotList.tsx
```diff
@@ -0,0 +1,228 @@
...
+        function handlePlotChange(changes: Partial<TemplatePlotConfig>) {
+            const newPlot = makeTrackPlot({
+                ...props.plot,
+                ...changes,
+            });
+
+            onPlotUpdate(newPlot);
+        },
+        [props.plot, onPlotUpdate]
+    );
```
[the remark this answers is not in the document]
Anders2303: I dont mind changing it back, but I'd argue that it's fine to just wrap callbacks by default, even if it doesn't do anything. React docs say that there's [no significant harm in doing so](https://react.dev/reference/react/useCallback#should-you-add-usecallback-everywhere), so I'd rather just wrap it in a `callback` as I'm writing, instead of needing to stop and think about whether they're _actually_ gonna be necessary in the end. 


Additionally, it might need to be wrapped in some future update (adding an effect, extracting the menu elements to a sub-component, etc...), so keeping callbacks wrapped can be convenient down the line.

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/SortablePlotList.tsx
```diff
@@ -0,0 +1,228 @@
...
+        },
+        [onUpdatePlots, props.plots]
+    );
+
+    const removePlot = React.useCallback(
+        function removePlot(plot: TemplatePlotConfig) {
+            onUpdatePlots(props.plots.filter((p) => p._id !== plot._id));
+        },
+        [onUpdatePlots, props.plots]
+    );
```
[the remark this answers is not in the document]
Anders2303: _[See my comment above](https://github.com/equinor/webviz/pull/794#discussion_r1863129177)_

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/TrackSettings.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    const curveHeadersQuery = useAtomValue(wellLogCurveHeadersQueryAtom);
+    const curveHeadersErrorStatus = usePropagateApiErrorToStatusWriter(curveHeadersQuery, props.statusWriter) ?? "";
+
+    const updateTrackConfig = React.useCallback(
+        function updateTrackConfig(configChanges: ConfigChanges) {
+            onUpdateTrack({ ...props.trackConfig, ...configChanges });
+        },
+        [props.trackConfig, onUpdateTrack]
+    );
```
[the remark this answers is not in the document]
Anders2303: _[See my comment above](https://github.com/equinor/webviz/pull/794#discussion_r1863129177)_

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/TrackSettings.tsx
```diff
@@ -0,0 +1,83 @@
...
+            <Dropdown
+                id="trackScale"
+                value={props.trackConfig.scale ?? ""}
+                options={PLOT_SCALE_OPTIONS}
+                filter={false}
+                onChange={(val) => {
+                    if (!val) updateTrackConfig({ scale: undefined });
+                    else updateTrackConfig({ scale: val as TemplatePlotScaleTypes });
+                }}
+            />
```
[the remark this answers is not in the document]
Anders2303: It was written before the dropdown supported general types! Just had to update the option array type, and the correct type is now inferred! 👍

frontend/src/modules/WellLogViewer/settings/settings.tsx
```diff
@@ -0,0 +1,120 @@
...
+        syncHelper.publishValue(SyncSettingKey.INTERSECTION, "global.syncValue.intersection", {
+            type: IntersectionType.WELLBORE,
+            uuid: wellboreUuid ?? "",
+        });
+    }
+    // Leave AFTER checking global, othwise the select menu will highlight the wrong value
+    const selectedWellboreHeader = useAtomValue(selectedWellboreAtom);
+
+    return [selectedWellboreHeader, setSelectedWellboreHeader];
+}
```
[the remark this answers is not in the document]
Anders2303: Issue created, see https://github.com/equinor/webviz/issues/817

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/SortableTrackItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <ListItemEndAdornment
+            track={props.trackConfig}
+            isExpanded={isExpanded}
+            onDeleteTrack={() => props.onDeleteTrack(props.trackConfig)}
+            toggleExpanded={() => setIsExpanded(!isExpanded)}
+        />
+    );
+
+    return (
+        <SortableListItem id={props.trackConfig._id} title={props.trackConfig.title} endAdornment={itemEndAdornment}>
```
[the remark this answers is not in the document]
Anders2303: I'll make a note for this for the next PR

frontend/src/modules/WellLogViewer/view/SubsurfaceLogViewerWrapper.tsx
Anders2303: _See above respone_

frontend/src/modules/WellLogViewer/view/SubsurfaceLogViewerWrapper.tsx
Anders2303: Could be nice. Leaving the css here for now, but added an issue in the component library git; https://github.com/equinor/webviz-subsurface-components/issues/2390

frontend/src/modules/WellLogViewer/view/ReadoutWrapper.tsx
Anders2303: Is react mounting expensive? The readout will toggle between show/hide whenever you move the mouse in and out of the log-viewer; is it better to mount/unmount it like I do now, is is the hide prop "more" correct for something that might rapidly show/hide

(Considering the size of the component, I assume the difference is negligible, so mostly just curious what the standard is

frontend/src/modules/WellLogViewer/settings/components/ViewerSettings/index.tsx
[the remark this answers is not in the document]
Anders2303: Issue created, see #818

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+        },
+        [setTrackConfigs, trackConfigs]
+    );
+
+    const handleTrackMove = React.useCallback(
+        function handleTrackMove(
+            movedItemId: string,
+            originId: string | null,
+            destinationId: string | null,
+            newPosition: number
```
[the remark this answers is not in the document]
Anders2303: Issue created, see #818

backend_py/primary/primary/services/ssdl_access/types.py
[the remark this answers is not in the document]
rubenthoms: A comment is good enough :+1:

frontend/src/modules/WellLogViewer/settings/atoms/derivedAtoms.ts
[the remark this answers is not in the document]
rubenthoms: My mistake, I assumed you selected all ensembles for the given field identifier - misread the `find` for a `filter` due to expectation. That's also why my second comment is not quite right. However, you are always using the first ensemble for the selected field identifier. What if you are having multiple ensembles for the same field? You are using the ensemble's case uuid to fetch stratigraphy. Apparently, we cannot assume that stratigraphy is constant among all cases (why would it otherwise be a parameter in the request). Should it be possible for the user to switch the case?

frontend/src/modules/WellLogViewer/settings/atoms/derivedAtoms.ts
```diff
@@ -0,0 +1,107 @@
...
+    const ensembleSetArr = get(EnsembleSetAtom).getEnsembleArr();
+    const selectedFieldId = get(userSelectedFieldIdentifierAtom);
+
+    if (ensembleSetArr.length < 1) {
+        return null;
+    } else {
+        const selectedEnsemble = ensembleSetArr.find((e) => e.getFieldIdentifier() === selectedFieldId);
+
+        return selectedEnsemble ?? ensembleSetArr[0];
```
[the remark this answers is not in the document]
rubenthoms: See my comment above.

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/SortablePlotList.tsx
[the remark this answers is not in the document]
rubenthoms: I beg to differ. In my opinion, there are some good reasons for not prematurely wrapping all functions in a `useCallback`.

- Unnecessary premature optimization/memory usage (latter admittedly having a rather small impact I assume)
- Depending on how the function is consumed down the line it might actually make the application less performant (might cost more CPU power than not having it, the function is created nonetheless, but you add an additional check) - it only leads to an additional re-render of the descendants if any of them has it somewhere in a dependency array (the descendants re-render anyways if the parent holding the callback re-renders, you can never stop that from happening - apart from memorizing the whole component of course)
- Unnecessary code and dependency arrays for each wrapped function that must be added and require maintenance - true that code might get refactored and true that these functions might get consumed in `useEffect`s but how often does that happen and is it too much to ask of a software engineer to double-check what he puts in a dependency array and/or sends down the line? (we actually also had some support for that in the `ESLint Exhaustive Deps rule` but I don't know if it's still there: https://dev.to/amitaldo/function-makes-the-dependencies-of-useeffect-hook-change-on-every-render-warning-in-react-35hd)
- Worsening readability as all functions get wrapped no matter if they eventually might end up in a dependency array or not (`wsc` is a good example for a codebase where almost everything is (used to be?) wrapped in `useCallback` and `useMemo`, no matter what - I personally find the code very hard to read)
- To break a fly on the wheel just in order to dodge comprehending code/possibly avoid future performance issues does not seem a very good argument to me
- I think you are actually presenting a good case against your statement with the code here as wrapping the functions does not add any benefit at all, just adds more code, and is not even consumed in any sub component (as they are not directly propagated). To argue that everything that is consumed in the HTML part should be wrapped does not make sense to me (or maybe I misunderstand what you're implying) as the HTML parts are not the problem, dependency arrays causing re-renders are. Hence, extracting the elements to a sub-component is per-se not a reason to wrap the functions - using them in a `useEffect` might be.

On the other hand, I admit that it's sometimes hard to know how a function gets consumed down the line, especially in components coming from other packages. In this latter case, I do fully agree that a `useCallback` can come handy. I am not against using `useCallback` per se, just against establishing a rule for wrapping everything in them. I definitely vote for using it with care since I value readability, cleanliness, and maintainability more than premature prevention. 

-IMO

frontend/src/modules/WellLogViewer/view/ReadoutWrapper.tsx
rubenthoms: Unfortunately, I can only give you the unsatisfactory answer: it depends.

Usually, it is not expensive to unmount/mount. However, it depends on the side-effects of a mount/unmount - basically, that what is happening in all the `useEffect`s in the respective component (and all its children). Moreover, you loose all your states of course. On the other hand, what is an alternative to toggling visibility via unmount/mount? You could of course make use of CSS properties (`display: hidden | height: 0` etc.) but all the elements will remain in the DOM-tree adding to its calculation-costs. I think the best practice I can recommend is to use mount/unmount as a default unless you have a good reason not to (mounting expenses and/or state). 

However, if you decide to keep the unmount logic here in the parent, you don't need the `hide` prop and could remove it.

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/private-components/SortablePlotList.tsx
```diff
@@ -0,0 +1,223 @@
...
+                        options={props.curveHeaderOptions}
+                        onChange={(v) => handlePlotChange({ _logAndName2: v, name2: v.split("::")[1] })}
+                    />
+                </>
+            )}
+            <div className="text-xs w-28 flex-shrink-0">
+                <Dropdown
+                    value={props.plot.type}
+                    options={PLOT_TYPE_OPTIONS}
+                    onChange={(v) => handlePlotChange({ type: v as TemplatePlotTypes })}
```
rubenthoms, line 170: `as TemplatePlotTypes` does not seem to be necessary.

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/private-components/SortablePlotList.tsx
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// It's my understanding that the STAT logs are the main curves users' would care about, so sorting them to the top first
+function sortStatLogsToTop(o: WellboreLogCurveHeader_api) {
+    if (o.logName.startsWith("STAT_")) return 0;
+    else return 1;
+}
+
+// The select value string needs a specific pattern
+type CurveDropdownOption = { value: TemplatePlotConfig["_logAndName"] } & DropdownOption;
```
rubenthoms, line 194:
```typescript
type CurveDropdownOption = DropdownOption<TemplatePlotConfig["_logAndName"]>;
```
?

frontend/src/modules/WellLogViewer/view/components/SubsurfaceLogViewerWrapper.tsx
```diff
@@ -53,30 +52,37 @@ export type SubsurfaceLogViewerWrapperProps = {
...
             }
+
+            return workbenchServices.subscribe("global.hoverMd", handleGlobalValueChange, instanceId);
         },
-        [wellLogController, wellboreUuid]
+        [instanceId, wellboreUuid, workbenchServices, wellLogController]
     );
+}
 
+function useCreateGlobalHoverMdBroadcast(
```
rubenthoms, line 81: Maybe `useCreateGlobalHoverMdBroadcastFunc(tion)`?

frontend/src/modules/WellLogViewer/view/components/SubsurfaceLogViewerWrapper.tsx
```diff
@@ -86,53 +92,53 @@ function useGloballySyncedMd(
...
+                instanceId
             );
 
             return unsubscribe;
         }
-    }, [workbenchServices, verticalSyncActive, syncableSettingKeys, moduleInstanceId, handleGlobalVertScaleChange]);
+    }, [workbenchServices, wellLogController, verticalSyncActive, syncableSettingKeys, instanceId]);
+}
+
+function useCreateGlobalVerticalScaleBroadcast(
```
rubenthoms, line 127: See above.

frontend/src/modules/WellLogViewer/view/components/SubsurfaceLogViewerWrapper.tsx
```diff
@@ -169,18 +175,29 @@ export function SubsurfaceLogViewerWrapper(props: SubsurfaceLogViewerWrapperProp
...
-    // Global value syncronization
-    const broadcastGlobalMdChange = useGloballySyncedMd(
-        props.wellboreHeader,
-        wellLogController,
+    const instanceId = props.moduleProps.viewContext.getInstanceIdString();
+    const syncableSettingKeys = props.moduleProps.viewContext.useSyncedSettingKeys();
+    const wellboreUuid = props.wellboreHeader?.wellboreUuid ?? "";
+
+    // Set up global hover md synchronization
+    useSubscribeToGlobalHoverMdChange(props.moduleProps.workbenchServices, wellLogController, instanceId, wellboreUuid);
```
rubenthoms, line 183: I think this is much more comprehensive now :smiley:

rubenthoms (approved):
Some remarks, otherwise good to go :+1:

frontend/src/modules/WellLogViewer/view/components/SubsurfaceLogViewerWrapper.tsx
Anders2303: 👍

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/private-components/SortablePlotList.tsx
Anders2303: Sure should be

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/private-components/SortablePlotList.tsx
Anders2303: Yes. Also could get rid of the one used on "addPlot" aswell, just needed to update the `AddItemButton` comp to pass the type

PR #813, opened 2024-11-26
[Group Tree / Flow Network] Refactor - Change use of "Group tree" to "Flow network"

Anders2303:
Renames a bunch, and moves some definitions and files to hopefully make the difference between Sumo's "Group Trees" and the webwiz rendered "group tree plot" more clear


jorgenherje (changes requested):
Nice work! A lot of nice adjustments with respect to separating into functions for cleanness and readability.

Some comments regarding naming, function attributes and usage of `pd.Series`

jorgenherje (approved):
Looks good to me!

I adjusted according to discussion

PR #823, opened 2024-12-06
Refactor: Make smda access use field ident only when neccessary

Anders2303:
Not all calls to SMDA actually requires the field identifier parameter, making it's use more cumbersome than it needs to be.
This PR updates the service (and relevant endpoints) to take field ident as a normal param when necessary instead.


HansKallekleiv (approved):
👍

PR #824, opened 2024-12-06
Refactor - Removed unused duplicated strat unit schema

Anders2303:
There was two `StratigraphicUnit` schemas. The one in the well-router was not used, so I removed it


HansKallekleiv (approved):
👍

PR #825, opened 2024-12-06
Refactoring - Added utility to simplify drogon identifier comparisons

Anders2303:
For all the "drogon or not" checks done in the router, we're manually writing out the identifier strings. This is error prone, and a bit cumbersome; I instead added small utility to keep these string checks in one place


HansKallekleiv (approved):
👍

PR #829, opened 2024-12-18
fix: `LayerManagerComponentWrapper` not updating when items in `LayerManager` are changing

rubenthoms:
Fixes issue with only having `SettingsGroup` available when adding items to a view as reported by user.

Also removes stored information in local storage when module gets removed now.


HansKallekleiv (approved):
👍

PR #841, opened 2025-01-07
browser cache

HansKallekleiv:


anders-kiaer:
I'm confident this will increase the perceived performance for users re-visiting the application. 👍 

I think we also want to give our endpoints the possibility to both define how long a cache is valid before a new response should be fetched from the frontend (`max-age`) but also how long a previously cached response can directly be shown to the user (`stale-while-revalidate`).

The way I understand how this flag can be used: E.g. for the (slow) well trajectory endpoint, we could have a `max-age` of e.g. one week and a `stale-while-revalidate` of many weeks. If the user revisits `webviz` after 10 days he/she will still get to see the trajectories immediately (since it less than `stale-while-revalidate`), and the browser will behind the scenes update the cache such that _next_ time the cache is visited it has a fresh instance.

https://web.dev/case-studies/ads-case-study-stale-while-revalidate

backend_py/primary/primary/config.py
```diff
@@ -29,7 +29,7 @@
     "ssdl": [SSDL_RESOURCE_SCOPE],
 }
 
-print(f"{RESOURCE_SCOPES_DICT=}")
-
+DEFAULT_CACHE_MAX_AGE = 3600  # 1 hour
+DEFAULT_STALE_WHILE_REVALIDATE = 3600  # 1 hour
```
anders-kiaer, line 33: Looks good. Just thinking loud, since most of the data we request/aggregate on is quite static (e.g. SMDA data, or calculations performed on a completed FMU run): Should we have the defaults to be `react-query-staleTime  < react-query-cacheTime <<  DEFAULT_CACHE_MAX_AGE << DEFAULT_STALE_WHILE_REVALIDATA`?

The "stale while revalidate" increases the time window where we have a browser cache hit (= where we immediately show data). At the same time, the user will not be stuck with "old" data in cache longer than max-age (shown to the user on next React component re-render/re-mount and timeout of React Query `staleTime` or page refresh). It would be interesting to see how this dynamic and potential upside turns out in practice, if the default is to have stale-while-revalidate larger than max-age.

backend_py/primary/primary/routers/explore/router.py
```diff
@@ -6,13 +6,15 @@
...
 from primary.services.utils.authenticated_user import AuthenticatedUser
+from primary.middleware.add_browser_cache import no_cache
 
 from . import schemas
 
 router = APIRouter()
 
 
 @router.get("/fields")
+@no_cache
```
anders-kiaer, line 17: This probably could be cached. It very rarely changes - and when it does the user has to wait for access to sync through infrastructure and Azure anyway. E.g. `max-age` of one minute and a default `stale-while-revalidate` of one day?

anders-kiaer (approved):
LGTM! This will be interesting to see the effect of in practice! ⏩

Issue #497, opened 2023-12-06
Create `m*n` grid layout for subplots in SimulationTimeSeries


jorgenherje:
Improve the control of the subplot grid layout for `SimulationTimeSeries`-module.

Rather than having one column with multiple rows, create `m*n` layout dependent on how many subplots are needed.

PR #848, opened 2025-01-15
Add subplot grid layout for SimulationTimeSeries

jorgenherje:
Add grid layout for subplots in `SimulationTimeSeries` module.

Make use of the `Figure` and `makeSubplots` utils for the module. Replaced the `subplotBuilder` with `PlotBuilder` and use same structure as other modules. Adjusted and added functionality for `Figure` and `makeSubplots`-util according to needs. Adjust existing modules according the adjustments/new features.

---
Closes: #497


anders-kiaer (approved):
LGTM from user perspective.

Anders2303 (approved):
Looks good to me as well on the logic side. Noted 3 very minor code quality things, but nothing important

frontend/src/modules/_shared/Figure.tsx
```diff
@@ -1,25 +1,59 @@
...
     private _plotData: Partial<PlotData>[];
     private _plotLayout: Partial<Layout>;
     private _gridAxesMapping: number[][];
+    private _axesToSubplotTitleAnnotationMap: Map<number, Partial<Annotations>>;
 
     constructor({
         data,
         layout,
         gridAxesMapping,
+        axesToSubplotTitleAnnotationMap: axesToSubplotTitleAnnotationMap,
```
rubenthoms, line 46: Maybe axis/axesIndexToSubplotTitleAnnotationMap?

frontend/src/modules/_shared/Figure.tsx
```diff
@@ -115,62 +202,132 @@ export class Figure {
...
             r: 0,
-            t: 0,
+            t: titleMargin,
             b: 0,
         },
+        title: options.title,
     };
 
-    const annotations: Partial<Annotations>[] = [];
+    const axesToSubplotTitleAnnotationMap: Map<number, Partial<Annotations>> = new Map();
```
rubenthoms, line 330: See above.

frontend/src/modules/_shared/Figure.tsx
```diff
@@ -273,14 +440,20 @@ export function makeSubplots(options: MakeSubplotOptions): Figure {
...
             }
 
-            if (options.subplotTitles && options.subplotTitles.length > index) {
-                const title = `<b>${options.subplotTitles[index]}</b>`;
-                annotations.push({
+            // If subplot titles are provided, add them as annotations.
+            // Set empty text if title is not provided for the subplot.
+            if (options.subplotTitles) {
+                const title =
+                    options.subplotTitles && options.subplotTitles !== true
```
rubenthoms, line 453: Another check if `options.subplotTitles` is defined does not seem necessary?

Also, should the titles maybe always be there and just set to "" when not given? This way, they might be updated later on in Figure.

rubenthoms (changes requested):
Nice work :+1: 
I mostly looked at the changes to the `Figures` class / file. A couple of minor comments, some naming questions.

frontend/src/modules/_shared/Figure.tsx
```diff
@@ -53,7 +91,13 @@ export class Figure {
         this._plotData.push(adjustedTrace);
     }
 
-    addAnnotation(annotation: Partial<Annotations>, row?: number, column?: number): void {
+    addShape(
+        shape: Partial<Shape>,
+        row?: number,
+        column?: number,
+        xCoordinateRef = CoordinateReference.DATA,
+        yCoordinateRef = CoordinateReference.DATA
```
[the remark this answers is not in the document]
jorgenherje: Not sure how often we want to change, but the idea is that one can add shapes for different usage. And have the possibility to add various shapes after construct for a subplot figure. Thereby the user has the freedom to add shapes both relative to the plot/subplot domain or a specific data value.

By using Domain, the placement of the shape will be kept, independent of zoom. This implies relative coordinates in the range [0, 1]. I.e. if you want a rectangle in the upper right corner, one can use domain coordinates. With coordinates set to data, it will follow actual (x, y)-coordinate values together with the data, e.g. if you would like to highlight at a specific timestamp or mark a specific y-value.

frontend/src/modules/_shared/Figure.tsx
```diff
@@ -86,21 +142,52 @@ export class Figure {
...
+        const adjustedAnnotation: Partial<Annotations> = {
+            ...annotation,
+            xref:
+                xCoordinateRef === CoordinateReference.DOMAIN
+                    ? (`x${axisIndex} domain` as XAxisName)
+                    : (`x${axisIndex}` as XAxisName),
+            yref:
+                yCoordinateRef === CoordinateReference.DOMAIN
+                    ? (`y${axisIndex} domain` as YAxisName)
+                    : (`y${axisIndex}` as YAxisName),
```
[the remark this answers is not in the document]
rubenthoms: Sounds good. I somehow missed that you have arguments for both x and y ref - two individual functions make sense.

frontend/src/modules/_shared/Figure.tsx
```diff
@@ -1,25 +1,59 @@
...
+import { Annotations, AxisType, Layout, PlotData, Shape, XAxisName, YAxisName } from "plotly.js";
+
+/**
+ * Enum for axis coordinate reference.
+ *
+ * DOMAIN: Use domain positions for the axis, a scalar/percentage position relative to the axis. Note this makes shape placed
+ *         same position in view independent of zoom.
+ * DATA: Use data for the axis, i.e. matching the value range of the axis. This it can follow zoom and axis range changes.
+ */
+export const enum CoordinateReference {
```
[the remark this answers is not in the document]
jorgenherje: Sure. I just followed the naming convention from plotly, and chose "data" for the one not specified as "domain". I'll change to "data domain" and "scene domain"

frontend/src/modules/_shared/Figure.tsx
jorgenherje: Changed from `subplotTitles?: true | string[]` to `subplotTitles?: string[]`, and always adding subplot title annotation. If `subplotTitles === undefined` or `subplotTitle[index]` is out of bounds, the text is set to empty string. Thereby annotations are always added when creating subplots.

frontend/src/modules/_shared/Figure.tsx
```diff
@@ -1,25 +1,59 @@
...
     private _plotData: Partial<PlotData>[];
     private _plotLayout: Partial<Layout>;
     private _gridAxesMapping: number[][];
+    private _axesIndexToSubplotTitleAnnotationMap: Map<number, Partial<Annotations>>;
 
     constructor({
         data,
         layout,
         gridAxesMapping,
+        axesIndexToSubplotTitleAnnotationMap: axesIndexToSubplotTitleAnnotationMap,
```
rubenthoms, line 46: Not necessary to assign to same variable name.

frontend/src/modules/_shared/Figure.tsx
```diff
@@ -273,38 +434,36 @@ export function makeSubplots(options: MakeSubplotOptions): Figure {
...
     }
 
-    if (options.subplotTitles) {
-        layout.annotations = annotations;
-    }
-
     return new Figure({
         layout,
         gridAxesMapping,
+        axesIndexToSubplotTitleAnnotationMap: axesIndexToSubplotTitleAnnotationMap,
```
rubenthoms, line 467: This is also not necessary when the variables and keys have the same name.

rubenthoms (approved):
Two minor comments, otherwise ready to be merged :+1:

PR #855, opened 2025-01-24
Well-log Viewer: Add discrete curves

Anders2303:
Upgrades the well-log viewer to also support discrete curves for geology and stratigraphy data from SMDA, along with various other improvements

### Features
* Upgrades `@webviz/well-log-viewer` to version `^2.1.0`
* Extends the existing `well/wellbore_log_curve_headers` and `well/log_curve_data` api endpoints to also return data from the mentionen SMDA APIs
  * The SMDA data is transformed to match the SSDL curve-header/curve-data payloads
* Adds menus for adding and modifying discrete curve tracks
* Changed the well-pick user flow to explicitly select stratigraphic columns and interpreters, instead of implying them from ensemble
  * This also removes the "unit/non-unit" pick terminology, showing only a single selection input instead
* Made some of the non-atomic query code for the module view to use proper atoms
* Various other smaller improvements and fixes


Anders2303:
One immediate question I have to the reviewer: currently, the multiple-source logic for the curve headers and data is currently being done within the router, with the data conversion being done inside the converter-file. Is this a bad setup? Should I rather move that logic into a service module, and get the source/conversion logic away from the router layer?

frontend/src/modules/WellLogViewer/utils/strings.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Simplifies some well log names to make them more suited for showing in the UI.
+ */
+export function simplifyLogName(logName: string, truncateLength = 0) {
+    // Removes curve-name, in case the log name is from a computed source (strat or geo)
+    logName = logName.split("::")[0];
+
+    // ! Unsure what sources there are, but this one seems to show up a lot for geology headers
+    if (logName === "OpenWorks R5000") return "OpenWorks";
+    // Fallback to ellipse for other long names
+    // TODO: Generalized shortening that doesn't truncate unless absolutely neccessary?
```
jorgenherje, line 30: As each if-statement returns, there is no need for `else if` and `else`, early return ensures no need for that.

if(...) return A;
if(...) return B;
return C;

frontend/src/modules/WellLogViewer/utils/strings.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export function getUniqueCurveNameForPlotConfig(plot: TemplatePlotConfig, nonUniqueNames?: Set<string>) {
+    if (!plot.name) throw new Error("Unexpected invalid config");
+    if (!plot._curveHeader) throw new Error("Unexpected invalid config");
+
+    if (nonUniqueNames?.has(plot.name)) {
+        return makeCompoundCurveName(plot._curveHeader.curveName, plot._curveHeader.logName);
+    } else {
+        return plot.name;
+    }
```
jorgenherje, line 44: No need for if and else, only if is needed

frontend/src/modules/WellLogViewer/utils/strings.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export function getUniqueCurveNameForCurveData(curve: WellboreLogCurveData_api, nonUniqueNames?: Set<string>) {
+    if (nonUniqueNames?.has(curve.name)) {
+        return makeCompoundCurveName(curve.name, curve.logName);
+    } else {
+        return curve.name;
```
jorgenherje, line 51: Here as well

frontend/src/modules/WellLogViewer/view/atoms/derivedAtoms.ts
```diff
@@ -26,3 +28,20 @@ function trajectoryToReferenceSystemPath(trajectory: WellboreTrajectory_api): nu
         return [easting, northing, tvd];
     });
 }
+
+// The Subsurface template pattern is kiiiinda stupid and only uses curve name both for data lookup and curve titles
```
jorgenherje, line 32: Perhaps remove "kiiiinda stupid"

frontend/src/modules/WellLogViewer/view/atoms/queryAtoms.ts
```diff
@@ -1,22 +1,61 @@
...
+            alt="test"
+        />
+    */
```
jorgenherje, line 39: Remove?

frontend/src/modules/WellLogViewer/view/atoms/queryAtoms.ts
```diff
@@ -1,22 +1,61 @@
...
+                    wellbore_uuid: wellboreUuid,
+                    curve_name: curveName,
+                    log_name: logName,
+                    source,
+                },
+            }),
+            enabled: Boolean(!locked && wellboreUuid && source && logName && curveName),
+        })),
+        combine(results: QueryObserverResult<WellboreLogCurveData_api, Error>[]) {
+            return mergeResults(results);
```
jorgenherje, line 58: Consider providing a specific type for the combine results for well log curve data, as commented for the headers

frontend/src/modules/WellLogViewer/view/view.tsx
```diff
@@ -72,18 +60,19 @@ export function View(props: ModuleViewProps<InterfaceTypes>) {
         );
     } else if (!mainElementsSuccess) {
         return <ContentError>Error loading curve data.</ContentError>;
+    } else if (!intersectionReferenceSystem) {
+        return <ContentError>Unexpected null for reference system.</ContentError>;
+    } else if (!wellboreTrajectoryDataQuery.data) {
+        return <ContentError>Unexpected null for trajectory data</ContentError>;
```
jorgenherje, line 66: Use if-statements instead of `else if` as this is early returns?

frontend/src/modules/_shared/components/ReadoutBox.tsx
```diff
@@ -93,7 +93,7 @@ export function ReadoutBox(props: ReadoutBoxProps): React.ReactNode {
     return (
         <div
             ref={readoutRoot}
-            className="absolute z-50 w-72 flex flex-col gap-2 p-2 text-sm rounded border border-neutral-300 bg-white bg-opacity-75 backdrop-blur-sm pointer-events-none"
+            className="absolute z-[9999] w-72 flex flex-col gap-2 p-2 text-sm rounded border border-neutral-300 bg-white bg-opacity-75 backdrop-blur-sm pointer-events-none"
```
jorgenherje, line 96: Should we have some sort of "standard" for things we want to have on top to prevent z-fighting? Or is it ok with these sort of hard coded high values?

jorgenherje (changes requested):
Nice work! Some comments.

Performed code review, not tested much.

frontend/src/modules/_shared/components/ReadoutBox.tsx
Anders2303: For this specific case, I'm figthing with the z-index of the well-log-viewer's overlay elements (the red hover-line has a high z-index). Didnt feel like figuring out the exact index i needed to go over it, so just went with the ol' reliable 9999

frontend/src/modules/WellLogViewer/view/atoms/derivedAtoms.ts
Anders2303: "exceptionally cumbersome" ?

frontend/src/modules/_shared/components/ReadoutBox.tsx
jorgenherje: Fair enough 👍

frontend/src/modules/WellLogViewer/settings/atoms/derivedAtoms.ts
```diff
@@ -1,46 +1,40 @@
...
+    // Selection-fixup. Default to field ident of first available ensemble if possible
+    if (!regularEnsembleArray.length) return null;
+    if (regularEnsembleArray.some((ens) => ens.getFieldIdentifier() === selectedFieldId)) return selectedFieldId;
 
-    return selectedEnsemble ?? regularEnsembleArray[0];
-});
-
-export const selectedFieldIdentifierAtom = atom((get) => {
-    return get(firstEnsembleInSelectedFieldAtom)?.getFieldIdentifier() ?? null;
+    return regularEnsembleArray[0].getFieldIdentifier();
```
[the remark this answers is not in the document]
Anders2303: I'll check if it was an easy fix. If not, I'll write this up as an issue, and save it for later

frontend/src/modules/WellLogViewer/settings/components/TemplateTrackSettings/ContinousTrackSettings.tsx
```diff
@@ -0,0 +1,60 @@
...
+                options={PLOT_SCALE_OPTIONS}
+                filter={false}
+                onChange={(val) => {
+                    if (!val) props.onFieldChange({ scale: undefined });
+                    else props.onFieldChange({ scale: val });
+                }}
+            />
+
+            <div className="col-span-2">
+                <PendingWrapper isPending={curveHeadersQuery.isPending} errorMessage={curveHeadersError ?? ""}>
```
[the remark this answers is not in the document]
Anders2303: I added "isPending". Sticking with that instead of "isLoading", as [the tanstack docs](https://tanstack.com/query/latest/docs/framework/react/guides/queries) use `isPending` in most examples.

frontend/src/modules/WellLogViewer/types.ts
```diff
@@ -1,23 +1,26 @@
...
 };
 
 /**
  * Extension of the SS-comp library type to add some state types to help with editing settings
  */
 export type TemplateTrackConfig = Omit<TemplateTrack, "plots"> & {
     // ID used to allow the settings-menu to drag-sort them
-    _id: string;
+    _key: string;
+    _type: WellLogCurveTypeEnum_api.CONTINUOUS | WellLogCurveTypeEnum_api.DISCRETE;
```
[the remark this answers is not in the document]
Anders2303: Since I've extended the `TemplateTrack` type from the log-viewer subsurface module, I wanted to make a clear distinction betwen the original fields, and the helper-fields I added. It'd be most convenient to keep it as is; although the usage of the types have grown a little past their initial version, so it might be worth making a dedicated "settings" type that's less directly dependent on the subsurface component... 🤔

frontend/src/modules/WellLogViewer/utils/logViewerTemplate.ts
```diff
@@ -1,100 +1,85 @@
...
+    config.name = config._curveHeader?.curveName;
+
+    // Reset config options that are only used in some specific cases
+    // config.color2 = undefined;
+    config.name2 = undefined;
+    config.fill = undefined;
+    config.fill2 = undefined;
+    config.colorMapFunctionName = undefined;
 
     switch (plot.type) {
```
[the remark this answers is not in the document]
Anders2303: I personally prefer switch statements for cases where I just need to check for specific strings; writing out `if (plot.type === "some string")` a bunch is annoying. Never found them any harder to read myself. Occasionally the fallthrough is useful. I thiiiiink I was decidedly using fallthrough (And I still am for the ") here while writing the update, but it stopped being relevant at some point...

For this one though, you're probably right, not really enough of a reason to use it here anymore, so Ill swap it to if statements

frontend/src/modules/WellLogViewer/utils/queries.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    // Data fetched, return and maybe apply transform
+    let data: T[] | K = _.map(results, "data") as T[];
+
+    if (data && dataTransform) {
+        data = dataTransform(data);
+    }
+
+    return { data: data as K, isLoading, isSuccess, isFetched } as UseQueryResult<K>;
+}
```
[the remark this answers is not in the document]
Anders2303: Went with `isPending`, [as recommended by tanstack docs](https://github.com/TanStack/query/discussions/6297)

jorgenherje (approved):
Nice work! Looks good to me 👍

PR #859, opened 2025-01-30
Sumo metadata - support grid geometry as geometry.name

HansKallekleiv:
Support the new fmu metadata where the name of a grid geometry in the grid property document is in `data.geometry.name`.
For now also support the old placement ( `data.tagname`)


anders-kiaer:
Not relevant anymore after changes in sumo/fmu-datiao.

PR #862, opened 2025-02-03
Use a single httpx client for each worker thread

HansKallekleiv:
SumoClient can now be initialized with existing httpx clients. https://github.com/equinor/sumo-wrapper-python/pull/227
This allows us to use a single client for the fastapi app context, thus removing time spent initializing such clients.


backend_py/primary/primary/services/sumo_access/_helpers.py
```diff
@@ -1,20 +1,49 @@
...
 
+class SynchronousMethodCallError(Exception):
+    """Custom error for when synchronous methods are called instead of async."""
+
+
+class FakeHTTPXClient:
+    """A fake HTTPX client to ensure we use async methods instead of sync ones."""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:  # pylint: disable=unused-argument
+        self._error_msg = "🚫 Do not use a synchronious http class!. Use the async http class instead. "
```
anders-kiaer, line 24:
```suggestion
        self._error_msg = "🚫 Do not use a synchronous http class!. Use the async http class instead. "
```

anders-kiaer (approved):
LGTM 👍 

Could consider adding a reference to https://github.com/equinor/fmu-sumo/issues/369 where the fake http client is defined.

Issue #777, opened 2024-10-29
Fix incorrect `Slider` value label position


jorgenherje:
Slider value label is positioned incorrectly when sliding through values. Jumps high when thumb is not on/near a specific tick/mark on the slider. When the mouse/thumb is near the tick, the label moves to correct position. Position is incorrect through slider movement.

----

See video. When no markers/ticks, the label is positioned incorrect all the time. If ticks/markers are available, the label is correctly places when thumb/mouse is on tick/maker:

https://github.com/user-attachments/assets/638c009c-41e3-4bf6-881b-b2b458d2abdc

Issue #645, opened 2024-06-13
`DiscreteSlider` slider ticks bug


jorgenherje:
When the slider has too many ticks, the slider line disappears. Only the draggable marker remains visible. Seems to occur when the number of steps becomes large.

Consider not showing ticks, but solid line if number of steps becomes too large?

Issue detected when testing `GroupTree`-module `Yearly` vs `Daily` sampling frequency

With 25 number of steps:
![image](https://github.com/equinor/webviz/assets/82032112/6400c152-d680-4878-a001-33874217f112)

Issue with 9053 number of steps:
![image](https://github.com/equinor/webviz/assets/82032112/b91911a9-e6be-4abd-8041-02119ff6abac)

PR #878, opened 2025-02-13
Adjust slider component

jorgenherje:
Fixed slider bugs.

Replaced label code, with slot named `valueLabel` for mui-base `Slider`-component. Render a `Popper`-component for the label.

Thereby there is no need to calculate positioning of label component anymore.

---

Closes: #777
Closes: #645


Anders2303 (approved):
Looks good!

PR #891, opened 2025-03-11
Framework/hover service

Anders2303:


frontend/src/framework/HoverService.ts
```diff
@@ -0,0 +1,195 @@
...
+
+export enum HoverTopic {
+    WELLBORE_MD = "hover.md",
+    WELLBORE = "hover.wellbore",
+    REALIZATION = "hover.realization",
+    TIMESTAMP = "hover.timestamp",
+    ZONE = "hover.zone",
+    REGION = "hover.region",
+    FACIES = "hover.facies",
+    WORLD_POS = "hover.world_pos",
```
jorgenherje: I think it would be good to clarify which type off coordinates we expect - e.g. utm-coordinates (i assume). Either do it in the naming of hover topic, or in the data below, i.e. `x` renamed to `xUtm`, and same for `y`. z is usually without suffix utm i believe.

frontend/src/framework/HoverService.ts
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /**
+     * Gets the current value for a given topic. If the requesting module is the one currently hovering, it will get the latest data immediately, otherwise it will get the latest throttled data.
+     * @param topic The HoverTopic to get the value for
+     * @param moduleInstanceId The id of the module requesting the value.
+     * @returns The latest or throttled value for the given topic
+     */
+    getTopicValue<T extends HoverTopic>(topic: T, moduleInstanceId: string): HoverData[T] | null {
+        // ? Should  this be an  opt-in functionality?
```
jorgenherje: Do you want to make it opt-in for the moduleInstance which is last hovered, or do you want it opt-in for modules even if they are not the last hovered?

I think current functionality seems like a good solution. We can rather make it opt-in when we see the need for it?

frontend/src/lib/utils/PublishSubscribeDelegate.ts
```diff
@@ -38,6 +38,16 @@ export class PublishSubscribeDelegate<TTopicPayloads extends TopicPayloads> {
...
 
+    subscribe(topic: keyof TTopicPayloads, handler: () => void): () => void {
+        const subscribers = this._subscribers.get(topic) || new Set();
+        subscribers.add(handler);
+        this._subscribers.set(topic, subscribers);
+
+        return () => {
+            subscribers.delete(handler);
+        };
+    }
```
jorgenherje: I think you can make `makeSubscriberFunction` below to use your new `subscribe` function.


```
makeSubscriberFunction(topic: keyof TTopicPayloads): (onStoreChangeCallback: () => void) => () => void {
        // Using arrow function in order to keep "this" in context
        const subscriber = (onStoreChangeCallback: () => void) => {
            return this.subscribe(topic, onStoreChangeCallback);
        };

        return subscriber;
    }
```

frontend/src/modules/_shared/DataProviderFramework/visualization/hooks/useSubscribedProviderHoverVisualizations.ts
```diff
@@ -17,93 +18,114 @@ export type AssemblerProviderHoverVisualizations<TTarget extends VisualizationTa
...
 function flattenVisualizationFunctionsRecursively<TTarget extends VisualizationTarget>(
     visualizationGroup: VisualizationGroup<TTarget> | AssemblerProduct<TTarget, any, any>,
 ): InternalAssemblerProviderHoverVisualizationFunctions<TTarget>[] {
     const visualizationFunctions: InternalAssemblerProviderHoverVisualizationFunctions<TTarget>[] = [];
-    if (visualizationGroup.hoverVisualizationFunctions) {
+    const groupId = Object.hasOwn(visualizationGroup, "id")
+        ? (visualizationGroup as VisualizationGroup<TTarget>).id
+        : undefined;
+
+    if (visualizationGroup.hoverVisualizationFunctions && groupId !== "") {
```
jorgenherje: Is it allowed with `groupId === undefined` if so, add comment for clarification. Otherwise add comment to state explicit checking for "valid" id string (i.e. non-empty)

frontend/src/modules/_shared/DataProviderFramework/visualization/VisualizationAssembler.ts
```diff
@@ -293,13 +285,15 @@ export class VisualizationAssembler<
...
             dataProviderManager.getGroupDelegate(),
+            [],
             options?.initialAccumulatedData ?? ({} as TAccumulatedData),
             options?.injectedData,
         );
     }
 
     private makeRecursively(
         groupDelegate: GroupDelegate,
+        inheritedDataProviders: DataProvider<any, any, any>[],
```
jorgenherje: As of now the inheritedDataProviders seems to be unused? Should it be removed?

jorgenherje (changes requested):
Really nice work! Well structured and easy to read code! Descriptive and easy to understand.

Some minor comments from my side.

frontend/src/framework/WorkbenchServices.ts
```diff
@@ -18,6 +18,7 @@ export type GlobalTopicDefinitions = {
     "global.infoMessage": string;
     "global.hoverRealization": { realization: number } | null;
     "global.hoverTimestamp": { timestampUtcMs: number } | null;
+    /** @deprecated use `global.hover.md` and `global.hover.wellbore` instead to delegate to hover-service */
     "global.hoverMd": { wellboreUuid: string; md: number } | null;
     "global.hoverZone": { zoneName: string } | null;
     "global.hoverRegion": { regionName: string } | null;
```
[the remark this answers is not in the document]
Anders2303: Made an issue [here](https://github.com/equinor/webviz/issues/1374)

frontend/src/modules/_shared/DataProviderFramework/visualization/VisualizationAssembler.ts
Anders2303: They are used; me and Ruben made it so the tree is flattened as you assemble the layers: a group will inherit all higher level providers. We use this parameter when we loop over children at line 370

frontend/src/modules/_shared/DataProviderFramework/visualization/VisualizationAssembler.ts
```diff
@@ -332,68 +329,85 @@ export class VisualizationAssembler<
...
             if (child instanceof DataProvider) {
-                numDataProviders++;
+                dataProviders.push(child);
+            }
+        }
 
-                if (child.getStatus() === DataProviderStatus.LOADING) {
-                    numLoadingDataProviders++;
-                }
+        for (const itemGroup of itemGroups) {
```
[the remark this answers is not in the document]
Anders2303: The inherited children logic forces us to do it like this for now: we need to collect all siblings on this level before we can recursively move to the next level. I think the impact is negligible, a user probably wont add enough layers for the extra  loop to matter

PR #896, opened 2025-03-21
Fix: Issues with element position tracking

rubenthoms:
Elements were not properly tracked anymore after recent changes. Introduced a new observation pattern utilizing the `IntersectionObserver`. We might want to look at the `Popover API` at some later point of time in order to remove custom code and to stick to the new standard: https://developer.mozilla.org/en-US/docs/Web/API/Popover_API

Also removed unused/unnecessary CSS classes.


Anders2303 (changes requested):

Anders2303 (approved):
Looks good too me!

PR #912, opened 2025-04-01
Callback function for defining actions for a group, variable titles and text, renaming of variables and files

rubenthoms:


Anders2303 (approved):
Looks good to me

PR #919, opened 2025-04-04
[DPF]: Renaming to `data provider` and refactoring of `VisualizationFactory`

rubenthoms:
- Renaming of `data layer` to `data provider` -> all occurrences of `layer` have been replaced with `provider`
- Refactoring of `VisualizationFactory` -> renamed to `VisualizationAssembler`, more generic logic supporting multiple encapsulated views
- Fixes issue with 2D viewer not being able to have more than one view


frontend/src/modules/_shared/DataProviderFramework/visualization/VisualizationAssembler.ts
```diff
@@ -0,0 +1,497 @@
...
+                for (const makeHoverVisualizationFunction of hoverVisualizationFunctions) {
+                    collectedHoverVisualizations.push(...makeHoverVisualizationFunction(hoverInfo));
+                }
+                return collectedHoverVisualizations;
+            },
+            customProps: {} as TCustomGroupProps,
+        };
+    }
+
+    private makeGroup<
```
Anders2303, line 364: This isn't something you need to fix for the merge, but the setup of the makeRecursive/makeLayer/makeGroup is bothering me a little bit.

`makeRecursively` as it stands now returns a... kinda finished group, with no ID or name. The initial makeRecursively is supposed to be making the "root" group, so shouldn't it just be handled the same as the normal group? As in: the main recursive call should just be a function called `makeGroupRecursively`, which returns a complete group, as opposed to calling recursive, THEN transforming that result and making a group from that.

As said thought, feel free to ignore this for **this** PR, but I think maybe that should be discussed, and maybe fixed later

frontend/src/modules/2DViewer/view/components/LayersWrapper.tsx
```diff
@@ -1,98 +1,126 @@
...
 
-const VISUALIZATION_FACTORY = new VisualizationFactory<VisualizationTarget.DECK_GL>();
+const VISUALIZATION_ASSEMBLER = new VisualizationAssembler<
+    VisualizationTarget.DECK_GL,
+    {
+        [GroupType.VIEW]: {
+            test: string;
+        };
+    }
+>();
```
Anders2303, line 67: This shouldnt be here, right?

frontend/src/modules/2DViewer/view/components/LayersWrapper.tsx
```diff
@@ -1,98 +1,126 @@
...
+VISUALIZATION_ASSEMBLER.registerDataProviderTransformers(
+    DataProviderType.DRILLED_WELL_TRAJECTORIES,
+    DrilledWellTrajectories,
+    {
+        transformToVisualization: makeDrilledWellTrajectoriesLayer,
+        transformToBoundingBox: makeDrilledWellTrajectoriesBoundingBox,
+    },
+);
+
+VISUALIZATION_ASSEMBLER.registerGroupCustomPropsCollector(GroupType.VIEW, View, ({ name }) => ({ test: name }));
```
Anders2303, line 123: also this

Anders2303 (changes requested):
Seems good to me. I'd only want to change the provider names, as discussed +  some leftover code I think you forgot to remove

Also wrote a remark about the make-function, but that can be left as a later issue

frontend/src/modules/_shared/DataProviderFramework/visualization/VisualizationAssembler.ts
rubenthoms: I think this requires some more discussion. A group is not necessarily a group that can be used in the view, it could also be a settings group whose children still need to be parsed recursively. A settings group, however, should not be returned as a group. There are other such groups as well (ItemGroup (hierarchy-wise) and Group (a group used as a view or similar) - we can discuss terminology). I think we best skip it for now as it does not change anything functionality-wise and take a discussion first.

PR #925, opened 2025-04-08
Snyk backend dependency updates

Anders2303:
Upgrades various dependencies to resolve issues reported by Snyk. Also did some assorted minor/patch updates while I was at it. Apart from some nullable schema-fields being generated as "optional", there shouldn't be any noticeable impact on the codebase. Any noteworthy changes are listed below

### Notable updates

 * `fastapi`: `1.103 -> 1.115`:
    * It appears that following this update, any schemas that defaults a field to None will have that respective field as optional in the typescript generated by generate-api, aka: `myField: str | None = None` generates `myField?: str | null` (as opposed to `myField: str | null`). 
    * As of v1.115, Pydantic models can now be used when defining queries, headers and such. See docs [here](https://fastapi.tiangolo.com/tutorial/request-form-models/).
    * v1.111 adds a [FastApi CLI](https://github.com/fastapi/fastapi/releases/tag/0.111.0), if we're interested.
    * Only breaking change is the use of `yield` in and except, which we don't use. If this affects any branch work you're doing, see [docs here](https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-with-yield/#dependencies-with-yield-and-except) and the changelogs for [v1.106](https://github.com/fastapi/fastapi/releases/tag/0.106.0) and [v1.110](https://github.com/fastapi/fastapi/releases/tag/0.110.0).
 * `pyarrow`: `12.0.1 -> 19.0.1`:
    * No breaking changes, just a bunch of new features and improvements
 * `httpx`: `0.24 -> 0.28`:
        * Client parameters `app` and `proxies` were removed, but I couldn't see that we were using them anywhere, so we should be fine

### Unresolved snyk-errors
Multiple issues reported by Snyk stem from the cryptography-library, which comes from the `msal` package. A refactor of the auth logic is in [the works](https://github.com/equinor/webviz/compare/main...sigurdp:webviz:refactor-auth) by @sigurdp , but the dependecy-update will have to wait for now. The `msal` package being stuck at `1.20`, means that the following packages cannot be updated:
  * `msal`: Version is pinned by us, see issue #ISSUE_NUMBER
  * `cryptography`: Snyk requires `>=43`, but `msal` pins it at `<41`
  * `azure-identity`: This goes the other way; newer versions require `msal > 1.20`

### Remaining outdated dependencies
There are still some dependencies that could (and probably should) be updated, but these are not related to any reported Snyk issues. Since these are major updates, they can likely have larger impacts across the code-base, and will therefore be worked on in their own dedicated PRs at some later time
```bash
astroid                              2.15.8   3.3.9    
black                                22.12.0  25.1.0   
msal                                 1.20.0   1.32.0   
numpy                                1.26.4   2.2.4    
pylint                               2.17.7   3.3.6    
python-dotenv                        0.21.1   1.1.0    
redis                                4.6.0    5.2.1    
uvicorn                              0.20.0   0.34.0   
xtgeo                                3.9.3    4.7.0    
```


anders-kiaer (approved):
LGTM 👍🏻

PR #930, opened 2025-04-09
Added explicity type for group prop collector args

Anders2303:


frontend/src/modules/_shared/DataProviderFramework/visualization/VisualizationAssembler.ts
```diff
@@ -90,17 +90,22 @@ export type VisualizationGroup<
     customProps: TCustomGroupProps[TGroupType];
 };
 
+export type GroupPropCollectorArgs<
```
rubenthoms, line 93: `GroupPropsCollectorArgs`?

rubenthoms (approved):
Looks good :+1: One missing `s`.

PR #932, opened 2025-04-09
[DPF]: Partial group type definitions in `VisualizationAssembler` and data provider type returned in assembler product

rubenthoms:


Anders2303 (approved):
Your changes look good. I added an additional field to the type aswell

PR #933, opened 2025-04-09
Refactored authentication and update MSAL version

sigurdp:
Refactored authentication to be able to use newer versions of MSAL library. MSAL upgraded to 1.32.0. Fixes #107 

Also improved performance of by caching the user's tokens in the server-side session store instead of creating a confidential client application and going via the token cache.

Other changes:
* Fixed `alive` and `alive_protected` endpoints
* Added `logout` endpoint
* Upgraded `cryptography` package
* Upgraded `azure-identity`


backend_py/primary/primary/routers/general.py
```diff
@@ -27,18 +27,25 @@ class UserInfo(BaseModel):
...
-
+async def get_alive() -> str:
     return f"ALIVE: Backend is alive at this time: {datetime.datetime.now()}"
 
 
 @router.get("/alive_protected")
 @no_cache
-def get_alive_protected() -> str:
-
+async def get_alive_protected() -> str:
```
Anders2303, line 36: Why are these marked as async if they dont do any awaits? Is this something particular about fast-api routes?

Anders2303 (approved):
LGTM

anders-kiaer (approved):
LGTM

frontend/src/framework/internal/components/LoginButton/loginButton.tsx
```diff
@@ -34,10 +34,7 @@ export const LoginButton: React.FC<LoginButtonProps> = (props) => {
     const { authState, userInfo } = useAuthProvider();
 
     function handleLogout() {
-        window.alert("Sign out not implemented in backend yet.");
-        /*
         window.location.href = "/api/logout";
-        */
```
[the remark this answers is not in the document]
sigurdp: I agree. 
Updated PR to use POST for the `logout` route and let the client redirect to "/" after logout

backend_py/primary/primary/routers/general.py
sigurdp: The current @no_cache decorator chokes on sync routes (which is the reason why in the current deployment these two routes fail.) Since we currently don't **need** the decorator in combination with sync routes, a simple work-around was to just make these two routes async

PR #941, opened 2025-04-23
fix: Adjust `SeismicAccess` after breaking change in `fmu-sumo==2.3.8`

rubenthoms:
`sas_token` and `url` are no longer exposed in `Cube` class in `fmu-sumo>=2.3.8`, see: https://github.com/equinor/fmu-sumo/pull/389/files#diff-89b6d98b408a75f18d17f2be7db80bfbda7c511403c0644411438bc9dfc86ae2

Pinning `fmu-sumo` to `2.3.8`.


sigurdp (approved):
👍

Issue #969, opened 2025-05-08
Temporary metadata workarounds


HansKallekleiv:
Checks and extra steps in our code due to maturing metadata.
All relevant code changes marked with `#969`
https://github.com/search?q=repo%3Aequinor%2Fwebviz%20%23969&type=code
Lets aim to close this before release.

- [ ] Checks needed for cases with old and new Sumo table aggregation
- [ ] Allowing volumetric tables with LICENSE column where the only value is Totals

PR #970, opened 2025-05-08
Hotfix: Fix parameter metadata search filter

HansKallekleiv:
Updated search context filter to pick up parameter tables from new Sumo cases.
Relates to #969


sigurdp (approved):
👍

PR #971, opened 2025-05-08
fix: WebGL context issues

rubenthoms:
This should fix the issue with `too many WebGL contexts` and the infinite loop.


jorgenherje (approved):
Looks good to me. Tested in separate branch 👍

PR #977, opened 2025-05-12
fix: insufficient key in template previews

rubenthoms:
Previews with multiple modules of same type cause duplicated key errors.


Anders2303 (approved):
LGTM

PR #984, opened 2025-05-14
Made open dropdowns close when clicking the collapse icon

### LINKED ISSUE #943: Fix ux when trying to close dropdown
When dropdown is opened, the list cannot be closed using the "expand/close" button.

This is probably an input filed with anchors, thereby the anchor has no action on click itself?

Click to expand:
Image

The close does not close the dropdown:
Image

Anders2303:
Resolves #943

### DIFF
--- a/frontend/src/lib/components/Dropdown/dropdown.tsx
+++ b/frontend/src/lib/components/Dropdown/dropdown.tsx
@@ -9,7 +9,6 @@ import { resolveClassNames } from "@lib/utils/resolveClassNames";
 import { convertRemToPixels } from "@lib/utils/screenUnitConversions";
 import { getTextWidthWithFont } from "@lib/utils/textSize";
 
-
 import type { BaseComponentProps } from "../BaseComponent";
 import { BaseComponent } from "../BaseComponent";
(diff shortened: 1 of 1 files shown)


frontend/src/lib/components/Dropdown/dropdown.tsx
```diff
@@ -540,7 +539,10 @@ function DropdownComponent<TValue = string>(props: DropdownProps<TValue>, ref: R
                             <IconButton
                                 size="small"
                                 className="align-sub"
-                                onClick={() => setDropdownVisible((prev) => !prev)}
+                                onClick={(evt) => {
+                                    if (dropdownVisible) evt.stopPropagation();
```
rubenthoms, line 543: I think we could actually always stop the event propagation.

Anders2303: We're using the built in label onclick to focus on the input field when you open the dropdown, so if we stop the propagation in general, we need to manually move focus over to the input

rubenthoms (approved):
LGTM

PR #985, opened 2025-05-14
Fix well router doc-strings

### LINKED ISSUE #826: Adjust comment for `getWellborePickIdentifiers`
After the recent changes, the field identifier seems no longer required for fetching wellbore pick identifiers - adjust the comment accordingly.

` Get wellbore pick identifiers for field and stratigraphic column` => ` Get wellbore pick identifiers for the given stratigraphic column`

Anders2303:
Resolves #826 and some other doc string errors I noticed


anders-kiaer:
@Anders2303 to check if still relevant.

PR #988, opened 2025-05-16
[Internal] Adds a simple internal utility for disabling the tanstack cache

Anders2303:
Adds a small utility to allow us to disable the tanstack cache by adding a flag to the local browser storage:
`webviz-debug:disable-tanstack-cache = true`


rubenthoms (changes requested):
We might want to consider a more fitting name than `tanstackCacheTime` as we are using it for both `gcTime` and `staleTime`. How about `debugOverrideTanstackTime` or `tanstackDebugTimeOverride`?

Anders2303:
> We might want to consider a more fitting name than `tanstackCacheTime` as we are using it for both `gcTime` and `staleTime`. How about `debugOverrideTanstackTime` or `tanstackDebugTimeOverride`?

I can go with `tanstackDebugTimeOverride`

rubenthoms (approved):
LGTM

PR #231, opened 2023-07-07
Plotly plot state wrapper

rubenthoms:
Stores and compares state of Plotly plots in order to prevent layout resets on data changes.

Closes #223.

Issue #223, opened 2023-06-28
Keep state in plotly plots when updating data


anders-kiaer:

anders-kiaer:
We will move away from plotly.

PR #1002, opened 2025-05-23
Framework - preserve plot state

Anders2303:
Creates a wrapper around `plotly-react` that should make ui state more stable. All this is doing is creating stable copies of provided data and layout objects, and injecting the `uirevision` prop to a fixed string if none is given (implementers can still trigger ui-resets by setting the prop manually, if needed).

Generally, this makes plots more stable, and partially solves #223, but there are still cases where where the ui ends up resetting, and it's hard to pinpoint exactly what causes it. Some causes seem to be:
* Many re-renders
* Data and layout changing during fetching, (`populated -> empty -> populated`)
  * I added a `plotUpdateReady` prop to stop the component from updating internally while fetching
* Some components are unmounting the component to show a loading icon, which (naturally) resets the state.
* Sometimes, it seemingly resets for no reason...

Generally, all plot components should be refactored to: keep the plotly comp mounted, memoize data/layout, use the new prop to avoid uneccessary internal state updates

### Future work
As mentioned, the uirevision approach is still very unstable. I think this band-aid solution is better than nothing, but we should probably consider a proper react-like ui state wrapper in the future, where we manually track ui changes and store them in the state, reapplying them each render (PR #231 presents a spike for this)

### DIFF
--- a/frontend/src/modules/ParameterCorrelationPlot/view/view.tsx
+++ b/frontend/src/modules/ParameterCorrelationPlot/view/view.tsx
@@ -15,6 +15,7 @@ import { useElementSize } from "@lib/hooks/useElementSize";
 import type { Size2D } from "@lib/utils/geometry";
 import { ContentInfo } from "@modules/_shared/components/ContentMessage";
 import { ContentWarning } from "@modules/_shared/components/ContentMessage/contentMessage";
+import { Plot } from "@modules/_shared/components/Plot";
 import type { ResponseData } from "@modules/_shared/rankParameter";
 import { createRankedParameterCorrelations } from "@modules/_shared/rankParameter";
 
@@ -46,7 +47,10 @@ export function View({ viewContext, workbenchSession, workbenchServices }: Modul
     const [prevParameterIdentString, setPrevParameterIdentString] = React.useState<string | null>(null);
 
     const syncedSettingKeys = viewContext.useSyncedSettingKeys();
-    const syncHelper = new SyncSettingsHelper(syncedSettingKeys, workbenchServices);
+    const syncHelper = React.useMemo(
+        () => new SyncSettingsHelper(syncedSettingKeys, workbenchServices),
+        [syncedSettingKeys, workbenchServices],
+    );
     const globalSyncedParameter = syncHelper.useValue(SyncSettingKey.PARAMETER, "global.syncValue.parameter");
 
     // Receive global string and update local state if different
@@ -59,15 +63,18 @@ export function View({ viewContext, workbenchSession, workbenchServices }: Modul
         [globalSyncedParameter, localParameterString],
     );
 
-    function handleClickInChart(e: PlotMouseEvent) {
-        const clickedPoint: PlotDatum = e.points[0];
-        if (!clickedPoint) {
-            return;
-        }
-        const newParameterString = clickedPoint.customdata as string;
-        syncHelper.publishValue(SyncSettingKey.PARAMETER, "global.syncValue.parameter", newParameterString);
-        setLocalParameterString(newParameterString);
-    }
+    const handleClickInChart = React.useCallback(
+        function handleClickInChart(e: PlotMouseEvent) {
+            const clickedPoint: PlotDatum = e.points[0];
+            if (!clickedPoint) {
+                return;
+            }
+            const newParameterString = clickedPoint.customdata as string;
+            syncHelper.publishValue(SyncSettingKey.PARAMETER, "global.syncValue.parameter", newParameterString);
+            setLocalParameterString(newParameterString);
+        },
+        [syncHelper],
+    );
 
     const numParams = viewContext.useSettingsToViewInterfaceValue("numParams");
     const corrCutOff = viewContext.useSettingsToViewInterfaceValue("corrCutOff");
@@ -189,7 +196,8 @@ export function View({ viewContext, workbenchSession, workbenchServices }: Modul
                     cellIndex++;
                 }
             }
-            setContent(figure.build(handleClickInChart));
+
+            setContent(<Plot data={figure.buildData()} layout={figure.buildLayout()} onClick={handleClickInChart} />);
             return;
         });
     }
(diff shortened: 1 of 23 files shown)


frontend/src/modules/SimulationTimeSeries/view/utils/PlotBuilder.ts
```diff
@@ -274,7 +272,11 @@ export class PlotBuilder {
             }
         }
 
-        return this._figure.makePlot({ onClick: handleOnClick });
+        return {
+            data: this._figure.makeData(),
+            layout: this._figure.makeLayout(),
+            ...customProps,
+        };
```
rubenthoms, line 279: Why changing the pattern? Either continue to use `makeLayout` and `makeData` as in the other builders or switch overall to the `makeProps` strategy. `customProps` seems unnecessary here?

frontend/src/modules/_shared/Figure.tsx
```diff
@@ -190,24 +192,6 @@ export class Figure {
         merge(this._plotLayout, patch);
     }
 
-    makePlot(plotArgs?: Omit<PlotParams, "data" | "layout">): React.ReactNode {
-        const config = plotArgs?.config ?? {
-            displaylogo: false,
-            responsive: true,
-            modeBarButtonsToRemove: ["toImage", "sendDataToCloud", "resetScale2d"],
-        };
```
rubenthoms, line 198: You moved the default config to your plot component, so it's not necessary to choose either strategy, but you should also here adhere to one: `makePlotProps` vs. `makeLayout/makeData`.

frontend/src/modules/_shared/components/Plot/plot.tsx
```diff
@@ -0,0 +1,79 @@
...
+    // By default, we try to keep the ui changes stable
+    uirevision: "revision_stable",
+};
+
+/**
+ * A wrapper utility of react-plotly that avoids the loss of UI state (zoom/pan/select)
+ * ! This component is still a bit unstable, and might reset if re-rendered multiple
+ * ! times in succession. Try to keep property references as stable as possible
+ * @param props
+ * @returns
```
rubenthoms, line 31: I don't see these last comment lines adding any useful information?

frontend/src/modules/_shared/components/Plot/plot.tsx
```diff
@@ -0,0 +1,79 @@
...
+    // Default to true if not defined
+    const shouldApplyPlotUpdate = plotUpdateReady ?? true;
+
+    // Store previous prop to avoid unnecessary rerenders
+    const [prevLayout, setPrevLayout] = React.useState<PlotParams["layout"]>(layout);
+    const [prevData, setPrevData] = React.useState<PlotParams["data"]>(data);
+    const [prevConfig, setPrevConfig] = React.useState<PlotParams["config"]>(config);
+
+    // ! Plotly seems to mutate objects given to it as props, so we need to clone them when
+    // ! For example, it changed the data's xAxis from "x1" to "x"
```
rubenthoms, line 45: Remove "when"?

frontend/src/modules/_shared/components/Plot/plot.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const [layoutStable, setLayoutStable] = React.useState<PlotParams["layout"]>(layout);
+    const [dataStable, setDataStable] = React.useState<PlotParams["data"]>(data);
+    const [configStable, setConfigStable] = React.useState<PlotParams["config"]>(config);
```
rubenthoms, line 48: Consider moving `stable` to the start, e.g. `stableLayout`. Otherwise, this could be confused with a question, hence, a boolean value.

frontend/src/modules/_shared/components/Plot/plot.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+    const [otherPropsStable, setOtherPropsStable] = React.useState<typeof otherProps>(otherProps);
```
rubenthoms, line 50: See above.

frontend/src/modules/_shared/components/Plot/plot.tsx
```diff
@@ -0,0 +1,79 @@
...
+        setPrevConfig(config);
+        setConfigStable(_.cloneDeep(config));
+    }
+
+    if (shouldApplyPlotUpdate && !_.isEqual(otherPropsStable, otherProps)) {
+        setOtherPropsStable(otherProps);
+    }
+
+    return React.useMemo(() => {
+        const layoutWithRevision = _.defaults({}, layoutStable, DEFAULT_LAYOUT);
```
rubenthoms, line 72: You have a minor discrepancy in the naming: You call it `DEFAULT_LAYOUT` as a general term, but here you are explicitly referring to its content. Consider either naming the variable here `layoutWithDefaults` or the `DEFAULT_LAYOUT` `LAYOUT_WITH_REVISION`. I would vote for the former.

frontend/src/modules/ParameterCorrelationPlot/view/view.tsx
```diff
@@ -46,7 +47,10 @@ export function View({ viewContext, workbenchSession, workbenchServices }: Modul
     const [prevParameterIdentString, setPrevParameterIdentString] = React.useState<string | null>(null);
 
     const syncedSettingKeys = viewContext.useSyncedSettingKeys();
-    const syncHelper = new SyncSettingsHelper(syncedSettingKeys, workbenchServices);
+    const syncHelper = React.useMemo(
+        () => new SyncSettingsHelper(syncedSettingKeys, workbenchServices),
+        [syncedSettingKeys, workbenchServices],
+    );
```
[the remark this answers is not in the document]
Anders2303: The status-writer is a dependency of the handleClickInChart callback below; I had to memoize it so the callback wasn't being rebuilt every render.

The general aim was to decrease the amount of re-renders on the Plotly component, as rapid re-rendering seemed to cause more ui resets...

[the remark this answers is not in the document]
rubenthoms: Ok, Thanks for the clarification. I'm fine with leaving it as-is. Alternatively and as long as `syncHelper` isn't used anywhere else, it could also be instantiated directly in the callback function to avoid the memoization. But both is okay in this case.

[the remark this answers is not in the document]
Anders2303: Ill leave it as is for now then

rubenthoms (approved):
LGTM :+1:

PR #1004, opened 2025-05-23
[DPF]: Controlled settings not notifying subscribers when static and getting controlled by an external setting

### LINKED ISSUE #1003: [DPF]: Investigate shared static setting value not applied to newly added providers
Instead, the default value seems to get applied.

Example:
Intersection with Extension length and adding a consuming provider (e.g. Seismic Intersection).

rubenthoms:
Fixes #1003.


jorgenherje (approved):
Looks good to me. Tested in Intersection Module branch

PR #1016, opened 2025-06-04
[DPF]: Add missing group properties in `VisualizationAssembler`

rubenthoms:


jorgenherje (approved):
Tested in my branch, and works well 👍

PR #1036, opened 2025-06-13
Framework - Automatic polling for updated ensembles

Anders2303:
* Adds timestamps to the ensembles we fetch from the backend.
* Adds a no-cached backend endpoint to look up the most recent timestamps
* Adds a hey-api plugin that allows us to add a "cache-busting" query parameter to requests, without the need to actually add it to the backend
* Make the workbench run a polling function every minute (could consider doing it more often) 
  * Whenever newer timestamps are found, the workbench will fetch the newest version in the background
  * New ensembles are fetched with the newest timestamp as a "cache-busting" query parameter


frontend/src/framework/Workbench.ts
```diff
@@ -291,6 +311,102 @@ export class Workbench {
         this._workbenchSession.setEnsembleSet(newEnsembleSet);
     }
 
+    private _pollingEnabled = false;
+    private _pollingInProgress = false;
+    private _waitingPollingRun?: NodeJS.Timeout;
+
+    beginEnsembleUpdatePolling(queryClient: QueryClient) {
+        // A polling is actively in progress, don't start a new one
+        if (this._pollingInProgress) return;
```
rubenthoms, line 320: Should this be `if (this._pollingEnabled)`?

frontend/src/framework/Workbench.ts
```diff
@@ -291,6 +311,102 @@ export class Workbench {
...
+        if (newSettings.length) {
+            this._updateExistingUserEnsembleSettings(queryClient, newSettings);
+        }
+
+        console.debug("checkForEnsembleUpdate - done...");
+
+        this._pollingInProgress = false;
+    }
+
+    private async _getLatestEnsembleTimestamps(
```
rubenthoms, line 373: See above.

frontend/src/framework/Workbench.ts
```diff
@@ -291,6 +311,102 @@ export class Workbench {
...
+        const timestamps = await queryClient.fetchQuery({
+            ...getTimestampsForEnsemblesOptions({ body: idents }),
+            staleTime: 0,
+            gcTime: 0,
+        });
+
+        return ensembles.map((ens, i) => [ens, timestamps[i]]);
+    }
+
+    private async _updateExistingUserEnsembleSettings(queryClient: QueryClient, newSettings: UserEnsembleSetting[]) {
```
rubenthoms, line 391: See above.

frontend/src/framework/Workbench.ts
Anders2303: No, it was intentional. The thought was to have further calls to the "begin" method restart the polling immediately, but I see that it might not be that easy to read.

Here, the `_pollingInProgress` specifically checks if the `pollForEnsembleChange` async function is actively running; if it is, we've basically already "restarted"

Anders2303: Since this case wasn't really used anywhere, I've just removed it, and gone with the more straightforward "_pollingEnabled" check instead.

PR #1039, opened 2025-06-16
Simplified debug flag string casing

Anders2303:
Made it camel-cased, and swapped the semi-colon with an underscore


sigurdp (approved):
👍

Issue #1043, opened 2025-06-23
Rename "Wellbore pick identifier" to "Wellbore pick interpreter" in Intersection-module


jorgenherje:
The setting named "Wellbore Pick Identifier" is not an identifier, but an interpreter.

Ref: https://github.com/equinor/webviz/pull/936#discussion_r2159040045

Issue #1035, opened 2025-06-13
[Intersection] Bad setState during in ViewportWrapper render


Anders2303:
> Warning: Cannot update a component (`DataProvidersWrapper`) while rendering a different component (`ViewportWrapper`). To locate the bad setState() call inside `ViewportWrapper`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
>> ViewportWrapper@http://localhost:8080/src/modules/Intersection/view/components/ViewportWrapper.tsx:47:35
>> DataProvidersWrapper@http://localhost:8080/src/modules/Intersection/view/components/DataProvidersWrapper.tsx:145:44
>> View@http://localhost:8080/src/modules/Intersection/view/view.tsx:35:31
>>...

The Intersection module is throwing the above error when it renders. Did some digging, and found that it's caused by `refocusViewport()` being called on the following lines:
```ts
// Line 116
if (props.doRefocus && props.focusBounds && isValidBounds(props.focusBounds)) {
    refocusViewport();
}

// ...

// Line 127
if (!isEqual(props.focusBounds, prevFocusBounds)) {
    setPrevFocusBounds(cloneDeep(props.focusBounds));

    // Update viewport if fit in view is ON
    if (props.focusBounds && fitInViewStatus === FitInViewStatus.ON) {
          refocusViewport();
    }
}
```

The refocus function fires a callback function back up to the parent, which in turn updates it's state. Moving the function calls to an effect should fix this

PR #1047, opened 2025-06-30
Intersection Module fixes

jorgenherje:
- Fix bad `setState()` in ViewportWrapper
- Rename Wellbore Pick Identifier setting -> SMDA Interpreter setting
- Set default vertical scale to 10.0

---

Closes: #1035
Closes: #1043


Anders2303 (approved):
Looks good 👍

PR #1058, opened 2025-07-08
Linter - Made the async suffix pylint plugin ignore builtin dunder-methods

Anders2303:
Noticed that the plugin was complaining about some builtin async functions, so made it avoid built-in python object functions


rubenthoms (approved):
I agree with Copilot that early returns are a bit more readable. Otherwise, good to go :+1:

PR #1066, opened 2025-07-31
Framework - Table Component rework

Anders2303:
**An example of the new table has been added to `MyModule2`**
image

Reworks the core table component to be easier to work with:
* Cleans up internals to be better separated and decoupled
* Supports externally controlled collation
* Supports custom rendering for column filters and data cells
* Supports skeleton rows for pending data
* Better typing of props so set-up is easier
* Changes the design to be more in line with Equinor design system
* Fixes various bugs

## Migration
Two components are currently using the existing table component: `InplaceVolumetricsTable` and `TornadoChart`. I've kept the old table around, but renamed it to `TableDeprecated` (and marked it with `@deprecated`). The two components should migrate over the new table, so we can delete the old one; the table props are mostly the same, except for the `tableHeading` object prop being swapped for a list of columns.
`tab`

## Future work
* Virtual columns
  * _Basically, data cells not directly taking a data field, which REQUIRES a render function_
* Alternating row colors
  * _A pretty standard thing in most list designs, so something we could consider._


frontend/src/lib/components/Table/table.tsx
```diff
@@ -1,648 +1,233 @@
...
+            if (!isLoadedDataRow<T>(row)) return true;
 
-    function handlePointerOver(row: TableRow<any> | null) {
-        if (props.onHover) {
-            props.onHover(row);
-        }
-    }
+            for (const columnId in tableFilterState) {
+                if (!(columnId in row)) {
+                    console.warn(`Attempting to filter data on column ${columnId}, which doesnt exist`);
```
jorgenherje, line 149: "doesnt" -> "doesn't"

frontend/src/lib/components/Table/table.tsx
```diff
@@ -1,648 +1,233 @@
...
-    }
+    const filteredRows = React.useMemo(() => {
+        if (props.controlledCollation) return rowsWithKey;
+        if (isEmpty(tableFilterState)) return rowsWithKey;
 
-    if (!isEqual(prevFlattenedHeadings, flattenedHeadings)) {
-        setPrevFlattenedHeadings(flattenedHeadings);
-    }
+        return rowsWithKey.filter((row) => {
+            // ? Unsure why it couldn't infer correctly here
```
jorgenherje, line 144: Not sure what you mean by this comment?

frontend/src/lib/components/Table/types.ts
```diff
@@ -0,0 +1,137 @@
...
+    render?: (props: ColumnFilterImplementationProps<any>) => React.ReactNode;
+    predicate?: (
+        filterValue: any,
+        dataValue: TData[TDataKey],
+        dataDef: DataCellDef<TData, TDataKey>,
+        entry: TData,
+    ) => boolean;
+};
+
+export type PendingData = { _pending: true };
```
jorgenherje, line 66: whis is it with underscore suffix? In usage we have to configure a pending row by setting `_pending: true` to actually render pending rows. In my head it should then be a regular public syntax?

frontend/src/lib/components/Table/types.ts
```diff
@@ -0,0 +1,137 @@
...
+    headerCells: HeaderCellDef[][];
+    filterCells: FilterCellDef<TData>[];
+    dataCells: { [TK in keyof TData]: DataCellDef<TData, TK> }[keyof TData][];
+};
+
+// --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- ---
+// - Table collation definitions - --- --- --- --- --- --- --- --- --- --- ---
+export enum SortDirection {
+    ASC = "asc",
+    DESC = "desc",
```
jorgenherje, line 113: Consider adding doc that "asc" and "desc" enum string values is chosen as this is used/passed to lodash `sortBy`-function in code?

frontend/src/lib/components/Table/utils.ts
```diff
@@ -0,0 +1,215 @@
...
+    return lowerValue.includes(filterString);
+}
+
+export function recursivelyBuildTableCellDefinitions<TData extends Record<string, any>>(
+    tableColumns: TableColumns<TData>,
+    depth: number = 0,
+    maxDepth = recursivelyCalcDepth(tableColumns),
+    // ! Object is mutated as the method runs
+    headerCells: HeaderCellDef[][] = [],
+): TableCellDefinitions<TData> {
```
jorgenherje, line 64: Consider creating an exported version with only `tableColumns`-prop. And have a private (non-exported) version which has the headerCells mutated?

Seems like we're only using it in `table-tsx`, and then only using `tableColumns`-prop.

frontend/src/lib/components/Table/utils.ts
```diff
@@ -0,0 +1,215 @@
...
+    const setValue = React.useCallback(
+        function setValue(newValue: TValue) {
+            if (useLocalValue) setLocalValue(newValue);
+            onValueChange?.(newValue);
+        },
+        [useLocalValue, onValueChange],
+    );
+
+    return [value, setValue];
+}
```
jorgenherje, line 215: Minor thought: Should this be placed in a file named `hooks.ts`? As we've usually done that for hooks elsewhere.

frontend/src/modules/MyModule2/view.tsx
```diff
@@ -1,44 +1,257 @@
...
+                alternatingColumnColors={alternateColColors}
+                selectable
+                multiSelect={allowMultiSelect}
+                // Listening to the internal changes, and make them update the controlled component
+                onSortingChange={setTableSortingState}
+                onFiltersChange={handleFilterUpdate}
+                onSelectedRowsChange={setSelectedRows}
+                onRowHover={setHoveredItem}
+            />
+             */}
```
jorgenherje, line 196: Remove?

jorgenherje (changes requested):
Really nice work! 👍

This is an initial code review from my side. I'll test the component in my own branch later on. Only nitpicks from my side.

Observations:
- Using multiselect with "shift", the text between first and second selected row becomes marked. I'd prefer "shift"+ select to be select range, or perhaps remove it? I.e. only use "ctrl" for multiselect?
image

- The input min/max for number range has low height.
image

frontend/src/lib/components/Table/table.tsx
Anders2303: I updated the the comment to clarify, but the problem is still there. Not really important to fix

Anders2303: Issue got fixed when I reworked the pending rows logic

frontend/src/lib/components/Table/private-components/tableRow.tsx
```diff
@@ -25,7 +25,7 @@ export function TableRow<T extends Record<string, any>>(props: TableRowProps<T>)
 
     return (
         <tr
-            className={resolveClassNames("group/tr border-b-2 last:border-b-0", {
+            className={resolveClassNames("group/tr border-b-2 last:border-b-0 select-none", {
```
jorgenherje, line 28: In my usage I find it handy to be able to copy some of the row data, but this disables possibility to mark and copy text. I assume this is the solution to prevent marking of text when using `shift`+`click` when performing multiselect of rows?
If so, is there another way to allow marking for copy, but prevent the text marking when multiselecting?

jorgenherje (changes requested):
Really nice work! 

Just one question/comment from my side. Otherwise all good 👍

jorgenherje (approved):

PR #1079, opened 2025-08-15
Patch: Fix wrong fullscreen state when using F11

### LINKED ISSUE #1074: [BUG]: Fullscreen button/logic does not work properly when switching to fullscreen with `F11`
When switching to full screen (whole app) by using the button in the right navigation bar, everything works as expected. However, when using `F11` to toggle full screen, the button's state does not get adjusted. Moreover, when clicking on the button now, nothing happens (the button state gets toggled, but the full-screen-mode is not exited). Exiting works with either `F11` or `ESC`, then the button state gets also properly updated.

Reproduce:
1) Press `F11`
2) Look at full screen button (wrong state)
3) Click on full screen button -> button state changes, nothing else happens
4) Click on button again -> button state changes back, nothing else happens

Anders2303:
Resolves #1074 

Chrome and Firefox had an issue were going into fullscreen via F11 wasn't correctly updating the state of the app's fullscreen button.

This is a known issue that doesn't seem to be prioritized ([moz bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=794468), so I added a workaround for now


frontend/src/framework/internal/hooks/useBrowserFullscreen.ts
```diff
@@ -19,13 +19,27 @@ export function useBrowserFullscreen(): [boolean, () => void] {
...
+        }
+
         function handleFullscreenChange() {
             setIsFullscreen(document.fullscreenElement != null);
         }
 
+        document.addEventListener("keydown", handleFullscreenKey);
         document.addEventListener("fullscreenchange", handleFullscreenChange);
-        return () => document.removeEventListener("fullscreenchange", handleFullscreenChange);
+        return function tearDownFullscreenListeners() {
```
rubenthoms, line 39: `removeFullscreenListeners`?

frontend/src/framework/internal/hooks/useBrowserFullscreen.ts
```diff
@@ -19,13 +19,27 @@ export function useBrowserFullscreen(): [boolean, () => void] {
...
+            document.removeEventListener("keydown", handleFullscreenKey);
+            document.removeEventListener("fullscreenchange", handleFullscreenChange);
+        };
     });
```
rubenthoms, line 43: Missing dependency array?

rubenthoms (changes requested):
Changes look good :+1: A couple of comments though. Also another question regarding the original implementation. You are only using `toggleFullscreen` in the effect and the outside world - it could handle entering and exiting fullscreen internally and we could get rid of the other two callbacks. Also some other considerations:
1) AFAIK Safari/macOS uses a different shortcut for full screen - however, since this is only a workaround, Safari should hopefully work correctly with the API
2) `toggleFullscreen` should be defined as function, not as arrow function, for debug purposes

rubenthoms (approved):
LGTM

PR #1082, opened 2025-08-19
[DPF]: Improve cancellation logic by tightening query fetches to Tanstack QueryObserver

rubenthoms:
This gives a more Tanstack-Query-native experience and fits better into the ecosystem. Cancellation is automatically performed by Tanstack when there are no more observers for a query. This plays well with `useQuery` hooks and `atomWithQuery/ies` as we let the QueryCache handle everything.


sigurdp (approved):
👍 Looks good and solves the immediate issue.

Hold back a little on merging this until we see that this PR is compatible with the changes needed to incorporate progress for long running operations

PR #1091, opened 2025-08-27
[Persistence]: System design and framework implementation/adjustment

rubenthoms:
This PR lays the foundation for persistence by defining the database structure, adjusting the framework, and implementing all required framework code.

- CosmosDB as database solution with emulator in dev environment
- Backend services and routers for database access / management of sessions and snapshots
- Adjustments of frontend framework
- New functionality in frontend framework to support persistence of workbench sessions and sharing of snapshots (business logic + components)


anders-kiaer:
Relevant input from today's work meeting: We should communicate clearly to the user that only _settings_ are part of the "snapshot" (use another name?), while data is based on what is available in Sumo at any given time.

Also a suggestion from Sigurd to later couple ensemble fingerprint to the snapshots in order to show e.g. warning when data has changed.

PR #1096, opened 2025-09-02
Tornado refactoring

### LINKED ISSUE #1078: Tornado/Simulation time series per sensitivity - Tornado bar sorting for multiple data sets
User wants to have the same sorting of bars in two tornado plots, e.g. when showing two ensembles.
A quick fix is to have the option to sort the bars alphabetically. This will be adequate for the user.

HansKallekleiv:
- Clearer separation of processing and visualization code
- Use new table component
- Various fixes, closes #1078


jorgenherje (changes requested):
Good work! :) Some comments from my side

frontend/src/modules/_shared/SensitivityProcessing/sensitivityProcessing.ts
```diff
@@ -0,0 +1,66 @@
...
+                    throw new Error(`Sensitivity type ${sensitivity.type} not supported`);
+            }
+        });
+}
+
+export const computeSensitivitiesForResponse = (
+    sensitivities: EnsembleSensitivities,
+    ensembleResponse: EnsembleScalarResponse,
+    referenceSensitivity: string,
+    barSortOrder: SensitivitySortOrder,
```
jorgenherje, line 41: This utility is used for both TornadoChart and table, thus the `SortBy` (if renamed) yields for more than bars in the TornadoChart

frontend/src/modules/TornadoChart/settings/settings.tsx
Anders2303, line 1: As a more general question; why **does** this module has a setting for  "view component"? I'd expect that a module called "Tornado Chart" can only be used to show... a tornado chart. Shouldn't the table variant just be a separate submodule (this would match how there's a table and plot module for Inplace Volumes)?

frontend/src/modules/TornadoChart/view/view.tsx
```diff
@@ -3,142 +3,137 @@ import React from "react";
...
-                    onSelectedSensitivity={setSelectedSensitivity}
-                />
+                <div className="w-full h-full" ref={wrapperDivRef}>
+                    <ContentInfo>No chart data available</ContentInfo>
+                </div>
             );
         }
+        return (
+            <div className="w-full h-full" ref={wrapperDivRef}>
+                <Plot layout={sensitivityChartBuilder.makePlotLayout()} data={sensitivityChartBuilder.makePlotData()} />
```
Anders2303, line 119: The layout seems a bit off, here, a lot of negative space on the left side, is that intentional?

image

Also, the bottom axis labels overlap with the plot title:
image

jorgenherje (approved):

Anders2303 (approved):

PR #1110, opened 2025-09-09
[DPF]: Persistence of group settings

rubenthoms:
Group settings were not persisted.


jorgenherje (approved):
Looks good to me!

Tested in my own branch, and it resolves my issue for persistence of group settings for the Intersection Module.

PR #1111, opened 2025-09-09
Adjustments to `SortableList` component

rubenthoms:
This PR makes the `SortableList` component more generic and makes it possible to use it with any custom components as items and groups.


frontend/src/modules/MyModule/view.tsx
```diff
@@ -1,11 +1,12 @@
 import React from "react";
 
-import type { PlotData } from "plotly.js";
+import { DragIndicator } from "@mui/icons-material";
+// import type { PlotData } from "plotly.js";
```
jorgenherje, line 4: Remove?

jorgenherje (approved):

PR #1144, opened 2025-09-25
Use POST requests for perforations, completions and casings from SubsurfaceDataApi

HansKallekleiv:
To facilitate webviz-4d functionality we switch to POST requests for this data.
This will allow us to fetch the data for multiple/all wells at once.

### DIFF
--- a/backend_py/primary/primary/services/ssdl_access/well_access.py
+++ b/backend_py/primary/primary/services/ssdl_access/well_access.py
@@ -46,7 +46,7 @@ class WellAccess:
     async def get_completions_for_wellbores_async(self, wellbore_uuids: List[str]) -> List[types.WellboreCompletion]:
         return await self._get_wellbore_list_data_async(
             wellbore_uuids=wellbore_uuids,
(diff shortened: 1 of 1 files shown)


backend_py/primary/primary/services/ssdl_access/_ssdl_request.py
```diff
@@ -0,0 +1,107 @@
...
+async def ssdl_post_request_async(
+    access_token: str,
+    endpoint: str,
+    data: Optional[List[str]] = None,
+    params: Optional[dict] = None,
+) -> dict:
+    """Convenience function for POST requests. Always returns a single dictionary."""
+    result = await _ssdl_request_async(access_token, endpoint, method="POST", data=data, params=params)
+    # POST requests always return dict, so we can safely cast
+    return result if isinstance(result, dict) else result[0]
```
Anders2303, line 34: You're not casting here? Is this comment outdated?

Anders2303 (approved):
Looks good to me, apart from some comments that could be changed

Issue #1001, opened 2025-05-22
Highlight that realization filter is active


HansKallekleiv:
Requested by user:
The realization filter is potentially dangerous. When activated, all further analysis includes it.
The user should continuously be aware that a realization filter is active.

anders-kiaer:
**First iteration:**
At least add symbol/icon in realization filter menu, when filter is active.
 
**Later(?) iterations:**
- Show same symbol in settings and/or module header.

Issue #991, opened 2025-05-19
Consider using TagPicker for multiple ensemble select?


jorgenherje:
When plotting/visualizing data from multiple ensembles, a lot of plotting/visualization might want to highlight/clarify order of selection when having multiselect.

The order of selection when using e.g. `EnsembleSelect`-component will allow the user to select which source to be plotted on top of the others. However it is not clear for the user that order of selection can be changed by reselecting in wanted order.

Consider using `TagPicker`-component instead, which clarifies order and multiselect. Following screenshot is from `InplaceVolumetricsTable`:

![Image](https://github.com/user-attachments/assets/858ce1ea-18d9-4acd-a032-be24d67ea425)

---

Additional functionality for the `TagPicker`: 
- Allow change of order after selection, e.g. by drag/drop of "tag" elements.

jorgenherje:
Example with `SimulationTimeSeries`-module. Difficult for user to understand that order of selection in "Ensembles" select-component is affecting the order in the view.
 
- When selecting "Drogon" first and "iter-1" second:
![Image](https://github.com/user-attachments/assets/d404773b-6c76-4033-b4a4-ca1a0619aa8f)


- When selecting "iter-1" first and "Drogon" second:

![Image](https://github.com/user-attachments/assets/aa7ee2cf-ec30-4419-8094-bca869463d5c)

PR #1155, opened 2025-09-30
[Library] Create `EnsemblePicker`-component

jorgenherje:
Create `EnsemblePicker` component using the `TagPicker`.

- Allows selecting ensembles in defined order
- Allows copy/paste of selection
- Visualize which ensemble has active realization filter

Added for usage in `SimulationTimeSeries`

----

Closes: #991
Partially closes: #1001


jorgenherje:
> Tested on `review` with `SimulationTimeSeries`. There seems to be an issue with deselecting/removing ensembles. I always have to click twice on either the `cross` or the `checkbox list` to remove a selected ensemble (see video). Might be related to the underlying tag picker component, in that case we should create an independent issue.
> 
>  Aufzeichnung.2025-10-03.095558.mp4

It seems like you're testing with only 1 ensemble. I've tested with multiple, and the issue you describe only happens when I deselect with 1 selected ensemble. I have a hunch that this actually is due to "fixup"-logic in the module settings/atoms rather than the `Picker`-functionality. I'll look into it.

PR #1158, opened 2025-10-01
Highlight active realization filtering in NavBar

jorgenherje:
Highlight that some realizations are bein filtered out due to the realization filter for given ensemble.

To ensure that user get information of the realization filtering affecting data visualized.

---

Closes: #1001


anders-kiaer (approved):
LGTM. For now, is it only `SimulationTimeSeires` that use this icon? Is the plan to change the other modules in follow-up PRs? :)

image

Issue #1192, opened 2025-10-15
Propagate more informative error message to frontend


anders-kiaer:
Primary backend log:
```
primary.services.service_exceptions.NoDataError: No per-realization tables found for: column_name='FGIR', table_name=None, content_type=['timeseries', 'simulationtimeseries'] [service=sumo]
INFO:     143.97.2.35:0 - "GET /timeseries/statistical_vector_data/?case_uuid=XXXXXXX&ensemble_name=iter-0&vector_name=FGIR&resampling_frequency=MONTHLY&realizations_encoded_as_uint_list_str=0-99 HTTP/1.0" 500 Internal Server Error
```

What the user sees in frontend log:

<img width="387" height="434" alt="Image" src="https://github.com/user-attachments/assets/256e0684-fdce-4da8-bb8e-1d8776fa8462" />

PR #1203, opened 2025-10-16
Propagate timeseries errors from backend to statuswriter

HansKallekleiv:
Relates to #1192


frontend/src/modules/SimulationTimeSeries/view/hooks/useMakeViewStatusWriterMessages.ts
```diff
@@ -5,11 +5,17 @@ import { EnsembleSetAtom } from "@framework/GlobalAtoms";
...
+import { queryIsFetchingAtom } from "../atoms/derivedAtoms";
+import {
+    vectorObservationsQueriesAtom,
+    regularEnsembleHistoricalVectorDataQueriesAtom,
+    vectorDataQueriesAtom,
+    vectorStatisticsQueriesAtom,
+} from "../atoms/queryAtoms";
 
 export function useMakeViewStatusWriterMessages(
     viewContext: ViewContext<Interfaces>,
```
jorgenherje: I know this was not introduced in your PR, but `viewContext` seems to be unused. Thus we should remove it.

PR #1206, opened 2025-10-17
Change Dialog to perform close from onPointerDown to onClick

jorgenherje:
Dialog closed with onPointerDown event, which prevents user to "regret" the mouse click and move mouse away from the close element.


sigurdp (approved):
👍

PR #1211, opened 2025-10-17
[Persistence]: Backend implementation

rubenthoms:
Adding:
- CosmosDB emulator as docker container and inital setup
- CosmosDatabase and CosmosContainer classes
- Stores for sessions, snapshots, and snapshot access logs
- Routers for persistence (sessions, snapshots, snapshot access logs)

Changing:
- Docker compose config: adding CosmosDB emulator container and adding startup dependency to primary container

### DIFF
--- a/backend_py/primary/primary/persistence/setup_local_database.py
+++ b/backend_py/primary/primary/persistence/setup_local_database.py
@@ -15,7 +15,6 @@ from primary.config import (
     COSMOS_DB_PROD_CONNECTION_STRING,
     COSMOS_DB_EMULATOR_URI,
     COSMOS_DB_EMULATOR_KEY,
-    COSMOS_DB_SKIP_SETUP,
 )
 
 LOGGER = logging.getLogger(__name__)


backend_py/primary/primary/persistence/cosmosdb/query_collation_options.py
Anders2303: This is not the latest version of the utility; the one in the management modal also has logic to handle filtered fields

Latest file is [here](https://github.com/Anders2303/webviz/blob/persistence/manager--management_modal/backend_py/primary/primary/services/database_access/query_collation_options.py)

backend_py/primary/primary/persistence/session_store/session_store.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+class SessionStore:
+    """
+    A simple data store for session documents with CRUD operations.
+    Supports pagination, sorting, filtering, and limits.
+    """
+
+    CONTAINER_NAME = "sessions"
+    DATABASE_NAME = "persistence"
```
sigurdp: These constants are only used internally in the class.
I suggest moving them to file/module scope as _DATABASE_NAME and _CONTAINER_NAME

sigurdp (approved):
More or less good to go.
Approved with a few questions/suggestions

Issue #1230, opened 2025-10-23
Standard result ´TagPicker` becomes to narrow in `SelectEnsemblesDialog`


jorgenherje:
When the browser gets too small width, the picker to select standard results becomes collapsed. Ref. `CaseExplorer` in the `SelectEnsemblesDialog`.

Solution:
- Set a min width to it, and consider making labels for the switch/toggle buttons shorter.
- Consider adding dynamic width calc for dialog - as already done for dialog height.

-----

<img width="1452" height="771" alt="Image" src="https://github.com/user-attachments/assets/7b13f74c-87d6-4e58-8507-f3e297690feb" />

PR #1234, opened 2025-10-23
Add dynamic width of SelectEnsemblesDialog

jorgenherje:
Minor adjustments to SelectEnsemblesDialog:

- Add dialog width calc to hook, as done for heigh
- Add optional max width/heigh props to Dialog-component
   - Optional props are set to `style` of div, and should override whatever set by tailwind css, i.e. `max-w-[75vw]` is overriden.
- Set standard result picker minWidth
- Reduce label text to reduce space

---

Closes: #1230


rubenthoms (approved):
Two minor comments, otherwise approved.

PR #1238, opened 2025-10-24
Patch: Tag input/picker fixes

Anders2303:
| image  | image |
|---|---|

Fixes assorted smaller issues in the `TagPicker`, `TagInput` and `EnsemblePicker` components
* Simplifies `TagInput` padding classes
* Exposes the tag-input's `inputProps` property in the `TagPicker` component
   * Tweaks the sizing of the input in `EnsemblePicker` to match tag height
* Adds a `dropdownMinWidth` prop to `TagPicker`. Defaults to 120px
* Fixes an issue where `TagPicker` would reset list focus too easily
* Added truncation of long un-wrappable tags in `TagInput`
* Hotfixes an issue where hovering at the start and end of the `TagPicker` dropdown would scroll
  * ~_I just tweaked the padding to make the list perfectly fit the amount of items. A more proper fix should separate hover and mouse focus movement but I want to wait with that until the hooks introduced in #1075 are available_~
* Made it easier to focus the `TagInput` (clicking the entire component root, instead of just the input field). The input field will also wrap around a little earlier now.


jorgenherje (approved):
Nice work!

PR #1258, opened 2025-10-29
[Dependencies] Update Subsurface component deps

### LINKED ISSUE #1135: Rotation/panning in 3D viewer fails with two views
How to reproduce:

1) Add two views:

Image

2) Try to rotate in the first view, resulting in both views going blank and an error message in console (`Uncaught TypeError: Cannot read properties of null (reading '0')`)

Image

Anders2303:
Updates each subsurface component dependency to their latest versions, introducing various bug-fixes and improvements. Most notably:
* Resolves #1135


HansKallekleiv (approved):
LGTM.
Can confirm sync issue is resolved.

PR #1263, opened 2025-10-29
[Persistence]: Implement persistence for `Pvt`-module

jorgenherje:
Implementation of module persistence for PVT


rubenthoms (approved):
LGTM! :+1: I committed a couple of changes to two persistable atoms and the settings component (using `SettingWrapper` now).

PR #1270, opened 2025-10-30
Bugfix: Disabled depth test for well-picks and trajectories in 2D viewer

Anders2303:
Surfaces are loaded in at Z-coordinate 0, so trajectories and wellpicks were clipping under them. This disables the depth test for the two to make sure they're always visible


rubenthoms (approved):
LGTM

PR #1283, opened 2025-11-01
[Persistence]: Implement persistence for `InplaceVolumesTable`-module 

HansKallekleiv:


jorgenherje (changes requested):
I get a crash when testing:

Have to debug further. Selected some settings, had two ensembles from same case. Then i chose single ensemble from new case

Image

jorgenherje (approved):
Fixed some issues. Looks good now

PR #1355, opened 2025-11-27
[DPF]: Make settings and available settings independent from each other

rubenthoms:


frontend/src/modules/_shared/DataProviderFramework/settings/implementations/DrilledWellboresSetting.tsx
```diff
@@ -8,60 +8,76 @@ import type {
...
+            return valueConstraints.map(mappingFunc);
+        }
+
+        return fixedValue;
+    }
+
+    isValueValid(currentValue: InternalValueType, valueConstraints: ValueConstraintsType): boolean {
+        function mappingFunc(value: WellboreHeader_api): string {
+            return value.wellboreUuid;
         }
```
jorgenherje: Same as prev comment regarding `mappingFunc`

frontend/src/modules/_shared/DataProviderFramework/settings/implementations/EnsembleSetting.tsx
```diff
@@ -9,14 +9,40 @@ import type {
...
+        return internalValue;
+    }
+
+    isValueValidStructure(value: unknown): value is ValueType {
+        if (value === null) {
+            return true;
+        }
+
+        // RegularEnsembleIdent is a class instance, check if it has expected methods
+        return typeof value === "object" && value !== null && "equals" in value && "toString" in value;
```
jorgenherje: Not very important, but as `equals` and `toString` are very common functions for a class, we could consider adding a bit more robust check? Consider having the current check, and thereafter retrieving the ident string using `toString()` and then validate it using `RegularEnsembleIdent.isValidEnsembleIdentString()`?

PR #1390, opened 2025-12-12
[4D]: New rich wellbore trajectory provider

rubenthoms:
Awaiting release of new wells layer in wsc. 

Depends on #1355.

### DIFF
--- a/frontend/src/modules/_shared/DataProviderFramework/settings/settingsDefinitions.ts
+++ b/frontend/src/modules/_shared/DataProviderFramework/settings/settingsDefinitions.ts
@@ -74,6 +74,17 @@ export enum Setting {
     FLOW_FILTER = "flowFilter",
 }
 
+enum ProductionPhase {
+    OIL = "oil",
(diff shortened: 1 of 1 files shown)


frontend/src/modules/_shared/DataProviderFramework/settings/implementations/PdmFilterSetting.tsx
```diff
@@ -0,0 +1,433 @@
...
+                    water: { value: 0, color: FLOW_COLORS.water_injection },
+                    gas: { value: 0, color: FLOW_COLORS.gas_injection },
+                },
+            };
+        }
+
+        return {
+            production: {
+                oil: {
+                    value: Math.min(currentValue.production.oil.value, 0),
```
Anders2303: The value is never negative, so this is always 0, right? 

I assume what you meant to do was this?
```ts
value: Math.min(currentValue.production.oil.value, valueContraints.production.oil),
```

PR #1397, opened 2025-12-16
Combine parameters and sensitivities into a single endpoint

### LINKED ISSUE #1396: Combine parameters and sensitivities into a single query
We retrieve parameters and sensitivities as two separate queries when loading ensembles.
These use data from the same underlying Sumo table. In some cases one of the queries may fail if both are triggered in fast succession.
The ensemble dialog will then give an error and the ensemble will not be loaded.

HansKallekleiv:
Closes #1396 

- Combine parameters and sensitivity data into a single endpoint to avoid querying Sumo for the same data twice.
- Remove unused endpoints
- Return an empty array for sensitivities instead of None. Frontend modules using sensitivity functionality already checks the length of the array.

### DIFF
--- a/backend_py/libs/services/src/webviz_services/sumo_access/parameter_access.py
+++ b/backend_py/libs/services/src/webviz_services/sumo_access/parameter_access.py
@@ -1,5 +1,4 @@
 import logging
-from typing import List, Optional, Tuple
 
 import pandas as pd
 import pyarrow as pa
@@ -34,7 +33,7 @@ class ParameterAccess:
         sumo_client = create_sumo_client(access_token)
         return cls(sumo_client=sumo_client, case_uuid=case_uuid, ensemble_name=ensemble_name)
 
-    async def get_parameters_and_sensitivities_async(self) -> Tuple[List[EnsembleParameter], List[EnsembleSensitivity]]:
+    async def get_parameters_and_sensitivities_async(self) -> tuple[list[EnsembleParameter], list[EnsembleSensitivity]]:
         """Retrieve parameters for an ensemble"""
         perf_metrics = PerfMetrics()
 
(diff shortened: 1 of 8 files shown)


backend_py/primary/primary/routers/parameters/router.py
```diff
@@ -8,93 +8,22 @@
 from webviz_services.sumo_access.parameter_types import EnsembleParameter, EnsembleSensitivity
```
jorgenherje: Unused imports (assume it will still be unused if you adjust according to previous comments)

frontend/src/framework/internal/EnsembleSetLoader.ts
```diff
@@ -357,20 +349,6 @@ async function loadEnsembleApiDataMapFromBackend(
             continue;
         }
```
jorgenherje: I know we did not check it previously, because the back-end failed if there was no parameters. But should we have a check for `parameterArray.length === 0` to ensure that the ensemble has non-empty parameter array?

jorgenherje (changes requested):
Looks good. Some minor comments from my side :)

backend_py/libs/services/src/webviz_services/sumo_access/parameter_access.py
```diff
@@ -1,5 +1,5 @@
 import logging
-from typing import List, Optional
+from typing import List, Optional, Tuple
```
jorgenherje: We can just use the built in `list` and `tuple`, which is best practice for the Python versions we use. 

We can also replace `Optional` with ` | None` notation, i.e. `str | None` for strings.

backend_py/primary/primary/routers/parameters/router.py
```diff
@@ -20,10 +19,11 @@ async def get_parameters_and_sensitivities(
...
-    parameters_and_sensitivities = await access.get_parameters_and_sensitivities_async()
-    return schemas.EnsembleParametersAndSensitivities(
-        parameters=converters.to_api_parameters(parameters_and_sensitivities.parameters),
-        sensitivities=converters.to_api_sensitivities(parameters_and_sensitivities.sensitivities),
+    parameters, sensitivities = await access.get_parameters_and_sensitivities_async()
+
+    return (
+        converters.to_api_parameters(parameters),
+        converters.to_api_sensitivities(sensitivities),
     )
```
jorgenherje: I actually  prefer the schema class `schemas.EnsembleParametersAndSensitivities` we had here for the api return type. I just felt it was unnecessary for the access-layer to wrap it in a new structure and access `parameters_and_sensitivities.parameters` and `parameters_and_sensitivities.sensitivities` immediately afterwards.

i.e. access layer has tuple[list[EnsembleParameter], list[EnsembleSensitivity]] and the router layer here returns 

```
return schemas.EnsembleParametersAndSensitivities(
   parameters=converters.to_api_parameters(parameters),
   sensitivities=converters.to_api_sensitivities(sensitivities),
)
```

PR #1403, opened 2025-12-29
[DPF]: Groups getting their own internal settings which can contribute via `valueRangeUpdater` to the intersection result of their shared settings

rubenthoms:
Depends on #1355.


jorgenherje (approved):
Approved with minor naming comments - replace "value range" naming to new "value constraints" naming.

Otherwise all good! Nice functionality or the Intersection-module! 👍

Issue #1416, opened 2026-01-13
[Bug]: SimulationTimeSeries module places query data in incorrect subplot


jorgenherje:
After https://github.com/equinor/webviz/pull/1371 was merged to main, the SimulationTimeSeries-module has bug due to `categorizedEnsembleIdentsAtom` in settings and `categorizedVectorSpecificationsAtom` in view. As the original requested vector specifications are provided in specific order, we loose this ordering after categorization.

Add sorting/assembling of query data in correct order in `vectorListQueriesAtom` in settings, and `vectorDataQueriesAtom` and `vectorStatisticsQueriesAtom` in view to make the `ensembleIdents`-array and the `vectorSpecifications`-array and their respective query results to be in the same order.

----

As the categorization is done within the `queryAtoms.ts`-files and never exported for external usage, the exported atoms should provide the query data in the same order as "external" use should expect. Thereby the exported atoms should assemble the query results sorted order. Thereby rest of usage can assume correct order.

Issue #1372, opened 2025-12-08
[Bug]: Fix `resamplingFrequency` selection in SimulationTimeSeries-module


jorgenherje:
FIx/clean up the selection of `resamplingFrequency` in the `SimulationTimeSeries`-module.

- `queryAtoms` in view: We should prevent/disable delta-vectors and statistics with flag - when having `resampleFrequency = undefined`?
- `view.tsx`: Provide a warning/error with the status writer if we request delta data when having `resampleFrequency = undefined`. 
- `settings.tsx`: We should probably also disable the "Raw(None)" option in "Resampling frequency"-dropdown in settings when requesting statistics data - if possible? 

Bug(s): 
- Seems to have issues when selecting "Raw" frequency when having delta ensembles. Setting states "Monthly", but query data is not montly.
- When selecting statistics and having a delta ensemble, there is no warning in the statusWriter to inform that statistics for delta is not supported.

---

Conclusion
- Disable queries for statistics (regular and delta ensemble) if resampling frequency is not valid, i.e. raw (provide error/warning message in ui)
- Disable queries (individual realization dta, and statistical data) for delta ensembles if resampling frequency is not valid, i.e. raw. (Add warning/error in statusWriter for the delta ensembles)

PR #1417, opened 2026-01-13
[Bug]: Fix SimulationTimeSeries issues with resampling frequency, and data in incorrect subplot

jorgenherje:
Fixes bugs/issues in SimulationTimeSeries-module:

- Data in incorrect subplot: Categorization of EnsembleIdents- and VectorSpecifications-array made the order of query result data to be inconsistent with the original arrays for fetching. Added sorting of query result according to original indices.

- Delta Ensembles cannot be used with Raw data. Drop fetching with "default" "Monthly" sampling, and rather disable queries if requesting with Raw. Added errors for statusWriter - as user should still be able to visualize regular ensembles with Raw data.

- Statistical summary data cannot be calculated with "Raw" data. As no statistics can be aggregated without specifying resampling frequency (both regular and delta ensembles), the view shows a `ContentWarning`stating invalid resampling frequency for statistical visualization.

---

Closes: #1372
Closes: #1416

### DIFF
--- a/frontend/src/modules/ModuleSerializedStateMap.ts
+++ b/frontend/src/modules/ModuleSerializedStateMap.ts
@@ -112,10 +112,6 @@ export type ModuleSerializedStateMap = {
     settings?: never,
     view?: never,
   },
-  "TopographicMap": {
-    settings?: never,
-    view?: never,
-  },
   "Vfp": {
     settings?: never,
     view?: never,
--- a/frontend/src/modules/ParameterResponseCorrelationMatrixPlot/interfaces.ts
+++ b/frontend/src/modules/ParameterResponseCorrelationMatrixPlot/interfaces.ts
@@ -2,10 +2,10 @@ import type { ParameterIdent } from "@framework/EnsembleParameters";
 import type { InterfaceInitialization } from "@framework/UniDirectionalModuleComponentsInterface";
 
 import {
-    parameterIdentsAtom,
     showLabelsAtom,
     useFixedColorRangeAtom,
     plotTypeAtom,
+    selectedParameterIdentsAtom,
     correlationThresholdAtom,
     hideIndividualCellsAtom,
     filterColumnsAtom,
@@ -29,7 +29,7 @@ export type Interfaces = {
 };
 
 export const settingsToViewInterfaceInitialization: InterfaceInitialization<SettingsToViewInterface> = {
-    parameterIdents: (get) => get(parameterIdentsAtom),
+    parameterIdents: (get) => get(selectedParameterIdentsAtom),
     showLabels: (get) => get(showLabelsAtom),
     useFixedColorRange: (get) => get(useFixedColorRangeAtom),
     plotType: (get) => get(plotTypeAtom),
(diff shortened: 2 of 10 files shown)


frontend/src/modules/SimulationTimeSeries/utils/querySortingUtils.ts
```diff
@@ -0,0 +1,58 @@
...
+ *     deltaEnsembleVectorSpecifications,
+ * ); *
+ * ```
+ */
+export function assembleQueriesInOriginalOrder<TQueryData, TRegularEnsembleItem, TDeltaEnsembleItem>(
+    regularEnsembleQueries: QueryObserverResult<TQueryData>[],
+    deltaEnsembleQueries: QueryObserverResult<TQueryData>[],
+    regularEnsembleCategorizedItems: CategorizedItem<TRegularEnsembleItem>[],
+    deltaEnsembleCategorizedItems: CategorizedItem<TDeltaEnsembleItem>[],
+): QueryObserverResult<TQueryData>[] {
```
Anders2303: Needing to set up two "categorizedItems" lists seems a bit cumbersome... wouldn't it be more clean to just keep a single list of "the combined order" (i.e. a list of keys) and just zipping/mapping them based on that? Idents are unique between regular- and delta-ensembles, right?

Something like:
```ts
export function zipQueriesInOrder(
  regularQueries: QueryResult[],
  deltaQueries: QueryResult[],
  orderingArr: str[], /* list of keys, such as ensemble stringified idents, or simply the key list directly from the node selector... */
): QueryResult[]
// ...
return [
  ...regularQueries,
  ...deltaQueries,
].sort((a, b) => {
   const keyA = ... /* whatever we need to do to turn this into a key */
   const keyB = ... /* whatever we need to do to turn this into a key  */

   return orderingArr.indexOf(identA) - orderingArr.indexOf(identB)
})
```

Unsure if this exact thing works cleanly with idents, but I think a generic and reusable version would look something like this

A fully generic version might be something like this?
```ts
export function zipInOrderBy(
  orderingArr: str[],
  makeKey: (item, index, sourceArr, sourceArrIndex) => key
  // Could match how lodash does stuff like this
  ...sourceArr[]
): any[];

File: frontend/src/modules/SimulationTimeSeries/settings/hooks/useMakeSettingAnnotations.ts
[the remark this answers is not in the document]
Comment (jorgenherje) [reply frontend/src/modules/SimulationTimeSeries/settings/hooks/useMakeSettingAnnotations.ts]: Renamed to `settingAnnotationHooks`

File: frontend/src/modules/SimulationTimeSeries/utils/querySortingUtils.ts (hunk 1 above)
Comment (jorgenherje) [reply frontend/src/modules/SimulationTimeSeries/utils/querySortingUtils.ts]: As discussed we must leave it as it is. Since `QueryObserverResult<TQueryData>[]` does not contain info about which query parameters the queries was performed with, we're not able to assemble back into order. Also I'll keep the regular and delta `CategorizedItems` as separate arrays, as they should match the order of the respective regular and delta `QueryObserverResult<TQueryData>[]`. 

I renamed the arguments to clarify that it is not "ensemble queries", but query results related to the categorized items:

```
export function assembleQueryResultsInOriginalOrder<TQueryData, TRegularEnsembleItem, TDeltaEnsembleItem>(
    regularEnsembleCategorizedQueryResults: QueryObserverResult<TQueryData>[],
    deltaEnsembleCategorizedQueryResults: QueryObserverResult<TQueryData>[],
    regularEnsembleCategorizedItems: CategorizedItem<TRegularEnsembleItem>[],
    deltaEnsembleCategorizedItems: CategorizedItem<TDeltaEnsembleItem>[],
): QueryObserverResult<TQueryData>[] {
   ...
```

Anders2303 (approved):

Issue #1266, opened 2025-10-30
[Framework]: Make debug modules available in production with enabled debug flag


rubenthoms:
- Move current deprecated modules to debug modules (for testing)
- Make debug modules available in production when debug flag enabled

PR #1440, opened 2026-01-26
[Framework]: Make debug modules available in production with enabled debug flag

jorgenherje:
Note: It does not include pure testing/debug modules specifically excluded in the `registerAllModules.ts`-file. `MyModule`, `MyModule2`, etc.

---

Closes: #1266


frontend/src/framework/internal/components/ModulesList/modulesList.tsx
```diff
@@ -449,67 +455,62 @@ export const ModulesList: React.FC<ModulesListProps> = (props) => {
...
+    const isVisible = drawerContent === RightDrawerContent.ModulesList;
+    const visibleModuleCategories = MODULE_CATEGORIES.filter(
+        (el) => el.category !== ModuleCategory.DEBUG || showDebugModules,
+    );
 
     return (
-        <div ref={ref} className={resolveClassNames("w-full h-full relative", { hidden: !visible })}>
+        <div ref={ref} className={resolveClassNames("w-full h-full relative", { hidden: !isVisible })}>
             <Drawer
-                visible
```
rubenthoms: The drawer could always remain visible as the wrapping div is already changing visibility.

rubenthoms (changes requested):
Some minor comments, otherwise LGTM

PR #1461, opened 2026-02-05
Support ert parameters

HansKallekleiv:


sigurdp (approved):
LGTM, discussed online on Teams
Old and new logic should probably be split into separate modules, but we decided to defer this split until later

Issue #1456, opened 2026-01-29
Reduce default width of module list


anders-kiaer:
When module list is opened, it takes up quite a bit of the dashboard space (first image below). It can be made smaller by the user and still have all module titles clearly visible (second image below), but the user selected with is not stored in localStorage for next time it is opened.

1) Either make a better default width and/or 2) remeber user preferred width of drawer?

<img width="2548" height="1268" alt="Image" src="https://github.com/user-attachments/assets/c13631a7-2d1b-4344-b32a-8aabde17c2ee" />

<img width="2550" height="1311" alt="Image" src="https://github.com/user-attachments/assets/7ec8dd9e-33b4-4839-a67a-dc42ee44afc2" />

Issue #1447, opened 2026-01-29
UX: Users confused about color panel not being possible to hide


anders-kiaer:
How to reproduce:
- Start new session.
- Click on color "Color session"
- This opens the color drawer, and also marks the icon with blue background in left menu.

A couple of user have asked why they can't close the color tab again by clicking the same icon.

<img width="624" height="613" alt="Image" src="https://github.com/user-attachments/assets/196ffac1-1ef3-4b76-9899-d4afdf123922" />

PR #1462, opened 2026-02-05
Improve Left/RightNavBar and Left/RightSettingsPanel

jorgenherje:
Improve UX for LeftNavBar and RightNavBar by separating actions related buttons in left, and content control to the right.

Separation of concerns: LeftNavBar opens dialogs, whilst RighNavBar controls content of RightSettingsPanel. The LeftSettingsPanel now always show settings for the active module instance.

Separation of concerns:
 - `LeftNavBar`: Open dialogs
 - `LeftSettingsPanel`: Module Settings (for active module)
 - `RightNavBar`: Control content of `RightSettingsPanel`
 - `RightSettingsPanel`: Show controls/settings for application based on selection in `RightNavBar`
     - Add Modules (control)
     - Module Log (list)
     - Realization Filter (control)

---

### Note

Made new `ReziableSettingsPanels`-component. This resembles the `ResizablePanels`-component, but works as a simplified version. Was created to have expand/collapse logic with wanted collapsed sizes for these panels. As the logic in `ResizablePanels` became unnecessary complicated when we had the need to control expand/collapse state externally, and calculate panels widths/heights based on collapse state changes, and react to `sizesInPercent` changes.

---

### Details

- LeftNavBar: Opens dialogs
   - SelectEnsemblesDialog
   - Templates Dialog

- LeftSettingsPanel now always shows settings for the active module instance.
  - Expand/collapse of module settings panel/drawer (never hidden all the way to zero width)
  - Switch between module settings/sync settings for the active module instance
  - Has a ModuleSettingsHeader which also controls show/dismiss of module warnings (also show warning banner).

- RightNavBar: Controls content of RightSettingsPanel
   - Click on buttons actively selects which settings to show in right settings panel
       - Add modules
       - Realization Filter
       - Module log
       - Color palette/settings

---

Closes: #1267
Closes: #1447 
Closes: #1456


frontend/src/framework/internal/components/LeftSettingsPanel/private-components/moduleSettingsHeader.tsx
```diff
@@ -0,0 +1,191 @@
...
+    activeModuleInstance: ModuleInstance<any, any> | undefined;
+    activeSetting: string | null;
+    availableSettings: Record<string, SettingConfig>;
+    isCollapsed: boolean;
+    onExpandClick: () => void;
+    onCollapseClick: () => void;
+    onSettingChange: (settingKey: string) => void;
+};
+
+export const ModuleSettingsHeader: React.FC<ModuleSettingsHeaderProps> = (props) => {
```
rubenthoms: See above.

frontend/src/framework/internal/components/LeftSettingsPanel/private-components/moduleSettingsHeader.tsx
```diff
@@ -0,0 +1,191 @@
...
+                {props.config.icon}
+            </button>
+        </Tooltip>
+    );
+};
+
+type ModuleSettingsHeaderProps = {
+    activeModuleInstance: ModuleInstance<any, any> | undefined;
+    activeSetting: string | null;
+    availableSettings: Record<string, SettingConfig>;
```
rubenthoms: I find the naming a bit confusing. There is a lot of use of the word "settings" all over the place - we should consider differentiating a bit better. Should we call "availableSettings" "availableSettingsTabs" and "activeSetting" "activeSettingTab - or remove the term "setting" entirely and just call them "availableTabs" and "activeTab" as the word "settings" is already in the context? Also consider adjusting the naming in the component.

frontend/src/lib/components/ResizableSettingsPanels/private-utils/sizeUtils.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Converts a size in pixels to a size in percent.
+ *
+ * Epsilon is added to avoid floating point precision issues when comparing sizes.
+ */
+export function pxToPercent(px: number, totalPx: number, epsilon?: number): number {
+    return (px / totalPx) * 100 + (epsilon ?? 0);
```
rubenthoms: This function seems pretty odd to me. I don't understand how this is supposed to fix floating point issues in the conversion. If you only want to compare two sizes, you should probably use something like fuzzyCompare (there is a util function in lib/utils/).  If the goal is to achieve a stable output, rather use something like this:

```
export function pxToPercent(
  px: number,
  totalPx: number,
  decimals = 4
): number {
  const factor = 10 ** decimals;
  return Math.round((px / totalPx) * 100 * factor) / factor;
}
```

frontend/src/lib/components/ResizableSettingsPanels/resizableSettingsPanels.tsx
```diff
@@ -0,0 +1,370 @@
...
+
+        function handlePointerUp() {
+            if (!dragging) return;
+            if (changedSizes.length > 0) {
+                storeConfigurationInLocalStorage(props.id, changedSizes);
+            }
+            dragging = false;
+            setIsDragging(false);
+            if (onSizesChange) {
+                onSizesChange(changedSizes);
```
rubenthoms: What if changedSizes was never set (i.e. when pointer down and immediately up again -> this would send [])?

rubenthoms (changes requested):
All in all some nice improvements. There are a couple of things that should/could be improved.

frontend/src/lib/components/ResizableSettingsPanels/resizableSettingsPanels.tsx
```diff
@@ -0,0 +1,370 @@
...
+                );
+
+                // Compute raw new width based on cursor position
+                // Left panel (index 0): width from container left edge to cursor
+                // Right panel (index 1): width from cursor to container right edge
+                const isRightPanel = index > 0;
+                const settingsPanelWidthPx = Math.max(
+                    isRightPanel ? containerRect.right - cursorX : cursorX - containerRect.left,
+                    0,
+                );
```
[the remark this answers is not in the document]
jorgenherje: Personally I prefer left and right

frontend/src/lib/components/ResizableSettingsPanels/private-utils/sizeUtils.ts
rubenthoms: I understand the use case. What feels odd to me is that this function is trying to handle multiple concerns at once, which makes its intent a bit unclear given the name.

Adding an epsilon can be useful for comparisons, but it doesn’t actually fix floating point precision - it biases the result. For a conversion utility, I’d prefer to keep it pure and apply tolerance at the comparison site (e.g. fuzzyCompare).

If the goal is stable values for display or persistence, rounding feels like a more explicit approach (e.g. pxToPercentRounded or roundToDecimals(pxToPercent(...))).

However, since you are removing it anyways, there is no reason for further discussion :)

frontend/src/lib/components/ResizableSettingsPanels/resizableSettingsPanels.tsx
```diff
@@ -0,0 +1,370 @@
...
+            return props.sizesInPercent;
+        }
+        const loadedSizes = loadConfigurationFromLocalStorage(props.id);
+        if (loadedSizes && loadedSizes.length === numSettingsPanels) {
+            return loadedSizes;
+        }
+        if (numSettingsPanels === 0) {
+            return [];
+        }
+        return Array(numSettingsPanels).fill(100.0 / (numSettingsPanels + 1));
```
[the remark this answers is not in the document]
jorgenherje: Not applicable anymore. Changed logic in https://github.com/equinor/webviz/pull/1462/changes/c8b422397defa33076dea3f45d454a63d7a7a5a0

frontend/src/framework/internal/components/LeftSettingsPanel/private-utils/useModuleWarning.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Hook to manage module warning visibility and dismissal.
+ */
+export function useModuleWarning(moduleInstance: ModuleInstance<any, any> | undefined) {
+    const storageKey = moduleInstance ? `module-warning-dismissed-${moduleInstance.getId()}` : null;
+    const warningText = useModuleWarningText(moduleInstance);
+
+    const [isDismissed, setIsDismissed] = React.useState(
+        () => !!storageKey && localStorage.getItem(storageKey) === "true",
```
rubenthoms: Nitpick: Is a function required for the computation of the initial value?

frontend/src/lib/components/ResizableSettingsPanels/resizableSettingsPanels.tsx
```diff
@@ -0,0 +1,391 @@
...
+ */
+export function ResizableSettingsPanels(props: ResizableSettingsPanelsProps): React.ReactNode {
+    const { onWidthsChange, onCollapsedChange } = props;
+
+    // Validate consistency of props
+    assertPropsConsistency(props);
+
+    function getInitialSettingPanelWidths(): SettingsPanelsWidth {
+        const defaultLeft = props.minWidths?.leftSettingsPanel ?? 30;
+        const defaultRight = props.minWidths?.rightSettingsPanel ?? 30;
```
rubenthoms: I think we should define `30` as a constant outside the component.

frontend/src/lib/components/ResizableSettingsPanels/resizableSettingsPanels.tsx
```diff
@@ -0,0 +1,391 @@
...
+                rightSettingsPanel: props.widthsInPercent.rightSettingsPanel ?? defaultRight,
+            };
+        }
+        return { leftSettingsPanel: defaultLeft, rightSettingsPanel: defaultRight };
+    }
+
+    const [isDragging, setIsDragging] = React.useState(false);
+    const [draggingPanel, setDraggingPanel] = React.useState<SettingsPanel>("leftSettingsPanel");
+    const [widths, setWidths] = React.useState<SettingsPanelsWidth>(getInitialSettingPanelWidths);
+    const [prevWidths, setPrevWidths] = React.useState(props.widthsInPercent);
```
rubenthoms: We should include the unit here as the default is px, but here we're using percent.

PR #1466, opened 2026-02-06
[DPF]: Added status writer implementation for dependencies

Anders2303:
image

Adds a status-writer object that's exposed at different parts of the DPF lifecycle, allowing user-facing information, warning, and error messages. Messages will show up on top of their respective data-provider or group

### DIFF
--- a/frontend/src/framework/GenericStatusMessageStore.ts
+++ b/frontend/src/framework/GenericStatusMessageStore.ts
@@ -1,14 +1,14 @@
 import { PublishSubscribeDelegate } from "@lib/utils/PublishSubscribeDelegate";
 
 import type {
-    PublishSubScribeStatusMessageStore,
+    PublishSubscribeStatusMessageStore,
     StatusMessage,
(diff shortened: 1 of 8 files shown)


Anders2303:
As discussed, we should look into having the messages show up at specific settings if possible (i.e, a dependency status message show up on each setting that depends on it), but that can be a future update

frontend/src/framework/GenericPubSubStatusWriter.ts
```diff
@@ -0,0 +1,74 @@
...
+    private _source: string | null = null;
+
+    constructor(source: string) {
+        this._source = source;
+    }
+
+    private addMessage(message: StatusMessage) {
+        this._messages.push(message);
+
+        this._pubSubDelegate.notifySubscribers(GenericStatusWriterTopic.UPDATE_MESSAGES);
```
rubenthoms: Should addMessage always immediately notify subscribers or should we debounce here?

PR #1495, opened 2026-02-23
Sessions - Error modals for session/snapshot persistence errors (load and save)

Anders2303:
* Added Error variant of Dialog component (red header, report issue button)
* Added red header variant of ConfirmationDialog component
* Added error modal popup for when snapshot and session persistence actions fail to replace error toasts
  * _Persistence recovery errors still use the Confirmation dialog; the logic flow for choosing retry/cancel/discard didn't fit that well with the save/load error modal. I did however make it so that confirmation modal uses the same red header_
* Made snapshot and metadata forms not using proper semantic form tag

### DIFF
--- a/frontend/src/framework/ConfirmationService.ts
+++ b/frontend/src/framework/ConfirmationService.ts
@@ -1,10 +1,10 @@
 export type ConfirmationVariant = "error" | "info";
-export type ConfirmActioncolor = "primary" | "danger" | "success" | "secondary";
+export type ConfirmActionColor = "primary" | "danger" | "success" | "secondary";
 
 export type ConfirmAction<T = string> = {
     id: T;
     label: string;
-    color?: ConfirmActioncolor;
+    color?: ConfirmActionColor;
 };
 
 export type ConfirmOptions<T = string> = {
--- a/frontend/src/framework/internal/components/SessionErrorDialog/sessionErrorDialog.tsx
+++ b/frontend/src/framework/internal/components/SessionErrorDialog/sessionErrorDialog.tsx
@@ -4,7 +4,7 @@ import { Check, Close, Error, Refresh } from "@mui/icons-material";
 import { useQuery } from "@tanstack/react-query";
 
 import { getAliveOptions } from "@api";
-import { GuiEvent, useGuiEventSubscriber, type GuiEventPayloads } from "@framework/GuiMessageBroker";
+import { GuiEvent, useRegisterGuiEventSubscriber, type GuiEventPayloads } from "@framework/GuiMessageBroker";
 import { SessionPersistenceAction } from "@framework/internal/WorkbenchSession/WorkbenchSessionManager";
 import type { Workbench } from "@framework/Workbench";
 import { Button } from "@lib/components/Button";
@@ -74,7 +74,7 @@ export function SessionErrorDialog(props: SessionErrorDialogProps): React.ReactN
     }
 
     // Subscribe to the session error gui-event
-    useGuiEventSubscriber(guiMessageBroker, GuiEvent.SessionPersistenceError, handleSessionSaveError);
+    useRegisterGuiEventSubscriber(guiMessageBroker, GuiEvent.SessionPersistenceError, handleSessionSaveError);
 
     return (
         <ErrorDialog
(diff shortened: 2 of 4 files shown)


frontend/src/framework/internal/components/GlobalConfirmationDialog/globalConfirmationDialog.tsx
```diff
@@ -30,6 +30,7 @@ export function GlobalConfirmationDialog(): React.ReactNode {
             modal
             showCloseCross={false}
             title={options.title}
+            variant={options.variant}
```
rubenthoms: See below.

frontend/src/framework/internal/WorkbenchSession/WorkbenchSessionManager.ts
```diff
@@ -602,8 +613,6 @@ export class WorkbenchSessionManager implements PublishSubscribe<WorkbenchSessio
             throw new Error(
                 "Could not load workbench session from data container. This should not happen and indicates a logic error.",
             );
-        } finally {
-            this._guiMessageBroker.setState(GuiState.SaveSessionDialogOpen, false);
         }
     }
```
rubenthoms: Maybe this function has grown to big and parts of it are not reusable. We could consider splitting it up into e.g.
- applySessionToGui(session)
- startPersistenceIfNeeded(session)
- startEnsemblePolling()
- publishSessionChanged()

frontend/src/lib/utils/errors.ts
```diff
@@ -0,0 +1,16 @@
+const REPO_URL = "https://github.com/equinor/webviz";
+
+export function reportErrorToGithub(error: Error, customStackReport?: string): void {
+    const stackToReport = customStackReport || error.stack || "";
```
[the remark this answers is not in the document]
Anders2303: Explicitly used `||` to avoid empty strings messages. I'll add a clarifying comment

Anders2303:
Created an overview issue for the comments regarding future work: https://github.com/equinor/webviz/issues/1504

frontend/src/framework/internal/components/SessionErrorDialog/sessionErrorDialog.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+    function onTryAgain() {
+        if (!errorEventPayload) return;
+
+        onCloseModal();
+        errorEventPayload.retry();
+    }
+
+    // Subscribe to the session error gui-event
+    useGuiEventSubscriber(guiMessageBroker, GuiEvent.SessionPersistenceError, handleSessionSaveError);
```
rubenthoms: Can we call the hook `useRegisterGuiEventSubscriber`? This is what the hook is actually doing, it is not using an already existing subscriber.

frontend/src/framework/ConfirmationService.ts
```diff
@@ -1,15 +1,17 @@
-import type { ButtonProps } from "@lib/components/Button/button";
+export type ConfirmationVariant = "error" | "info";
+export type ConfirmActioncolor = "primary" | "danger" | "success" | "secondary";
```
rubenthoms: Capital C

frontend/src/lib/components/Dialog/errorDialog.tsx
```diff
@@ -0,0 +1,91 @@
...
+export function ErrorDialog(props: ErrorDialogProps): React.ReactNode {
+    const isMountedRef = useIsMountedRef();
+    const runIdRef = React.useRef(0);
+
+    const [isSymbolicatingStack, setIsSymbolicatingStack] = React.useState(false);
+
+    async function maybeGetSymbolicatedTrace(error: Error) {
+        if (!shouldSymbolicate()) return undefined;
+
+        setIsSymbolicatingStack(true);
```
rubenthoms: Having the `isMount` check here is actually guarding against any added functionality in `reportIssue`. You are right that this function is not doing anything async before and in its current state, this should not become an issue. Adding the check is guarding us in the future should the functionality be expanded, especially in reportIssue (dev might not double-check what's inside this function).

PR #1512, opened 2026-02-26
Simplified Azure Monitor OpenTelemetry configuration in Radix

sigurdp:
This PR simplifies configuration of Azure Monitor telemetry by using workload identity to import Application Insights connection string secrets from Azure KeyVault, and then selecting the appropriate Insights connection string at runtime based on the currently running Radix environment.

It also fixes telemetry ingestion problems that were introduced in #1477.  Apparently version `1.8.6` of `webviz_core_utils.azure_monitor_destination` switched to using `RateLimitedSampler` in a patch version (https://github.com/Azure/azure-sdk-for-python/pull/44925). This resulted in partial telemetry samples and breaks our current usage of telemetry for debugging. Fixed by reverting to old behavoir of using fixed-rate sampling combined with pinning of the package version.

Changes:
* Changed how we configure the Application Insights connection string when running in Radix. We now use workload identity for secrets in both `backend-primary` and `user-grid3d-ri`. We import all connection string secrets as env variables and select the appropriate connection string at startup time in code.
* Added `webviz_core_utils.radix_utils` module with functions for Radix environment detection and metadata extraction.
* Refactored services to use `is_running_on_radix_platform()` for platform detection.
* Introduced `webviz_core_utils.azure_monitor_destination` with the `AzureMonitorDestination` dataclass to encapsulate telemetry destination logic, including environment-based connection string resolution and resource attributes.
* Refactored telemetry setup functions to use new utility methods.
* Fixed telemetry sampling when configuring Azure Monitor through `configure_azure_monitor()` by specifying `sampling_ratio=1.0`
* Pinned `azure-monitor-opentelemetry` to version `1.8.6`.
* Added support for secret/environment variable containing AppInsights connection string for `preprod` in `WEBVIZ-INSIGHTS-CONNECTIONSTRING-PREPROD`

### DIFF
--- a/frontend/src/framework/GuiMessageBroker.ts
+++ b/frontend/src/framework/GuiMessageBroker.ts
@@ -171,8 +171,6 @@ const persistentStates: GuiState[] = [
     GuiState.DevToolsVisible,
     GuiState.RightSettingsPanelWidthInPercent,
     GuiState.RightDrawerContent,
-    GuiState.NumberOfUnsavedRealizationFilters,
-    GuiState.NumberOfEffectiveRealizationFilters,
 ];
 
 export class GuiMessageBroker {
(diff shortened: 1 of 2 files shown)


anders-kiaer (approved):
One small comment - otherwise LGTM 👍

PR #1515, opened 2026-02-27
[Lib/DPF] Rework Menu component +Improve Action menu UX in DPF

Anders2303:
* Installs [`@base-ui/react`](https://base-ui.com/react/components/menu) as a new dependency
  * The library is now out of beta, so we can start to move away from `mui-base` (see #946 )
  * Reworks the `Menu` component in the `lib` to use the [menu-component from Base UI](https://base-ui.com/react/components/menu). Most notably, this involved removing the various menu-components that we used sporadically, and we now instead have a single component with menu entries defined via props. The removed components are:
    * `MenuButton`
    * `MenuDivider`
    * `MenuHeading`
    * `MenuItem`
    * `MenuText`
  * All components that used these are now properly using the shared menu component
* Added a `Popover` component in Lib, wrapping the [Base UI component](https://base-ui.com/react/components/popover)
  * The module event log button now uses this, as that seemed more correct than a full Menu component 
  * _The Help popovers that was introduced a little while ago should consider making use of this, but that'll be a separate PR_
* Also adds some smaller UX-improvments to DPF:
  * Base UI generally improves keyboard navigation for the menu
  * Items in the actions menu are now grouped in sub-menus for easier navigation
  * More confusing items can now have descriptions (short descriptor has been added View and ContextBoundary items)
  * When adding group items it's now possibly automatically open that item's action menu on mount
     * This makes it a bit quicker to create a view with content, and we think this should teach the user that you need to add layers to views using the views action-button

image
image


frontend/src/modules/2DViewer/settings/persistence.ts
```diff
@@ -22,7 +22,15 @@ const schemaBuilder = new SchemaBuilder<SerializedSettings>(() => ({
             nullable: true,
         },
         preferredViewLayout: {
-            enum: [PreferredViewLayout.VERTICAL, PreferredViewLayout.HORIZONTAL],
+            enum: [
+                ViewLayout.VERTICAL,
+                ViewLayout.HORIZONTAL,
+
+                // ! Legacy workaround
```
rubenthoms: Consider introducing a keyword that is easy to search for when starting on versioning.

frontend/src/modules/3DViewer/settings/persistence.ts
```diff
@@ -22,7 +22,15 @@ const schemaBuilder = new SchemaBuilder<SerializedSettings>(() => ({
             nullable: true,
         },
         preferredViewLayout: {
-            enum: [PreferredViewLayout.VERTICAL, PreferredViewLayout.HORIZONTAL],
+            enum: [
+                ViewLayout.VERTICAL,
+                ViewLayout.HORIZONTAL,
+
+                // ! Legacy workaround
```
rubenthoms: Look other comment.

frontend/src/modules/_shared/DataProviderFramework/ViewportLayoutMenu.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export type ViewportLayoutMenuProps = {
+    value: ViewLayout;
+    onValueChange: (newValue: ViewLayout) => void;
+};
+
+export function ViewportLayoutMenu(props: ViewportLayoutMenuProps): React.ReactNode {
+    function handleMenuItemClick(itemId: string) {
+        if (!Object.values(ViewLayout).includes(itemId as ViewLayout)) {
+            // throw new Error(`Invalid view layout: ${itemId}`);
```
rubenthoms: Legacy?

frontend/src/modules/_shared/DataProviderFramework/ViewportLayoutMenu.tsx
```diff
@@ -0,0 +1,62 @@
...
+            onActionClicked={handleMenuItemClick}
+            items={[
+                {
+                    id: "layout",
+                    label: "Preferred view layout",
+                    items: [
+                        {
+                            id: ViewLayout.HORIZONTAL,
+                            label: "Horizontal",
+                            checked: props.value === ViewLayout.HORIZONTAL,
```
rubenthoms: Maybe these should be radio instead?

rubenthoms (changes requested):
All in all, this looks very good :+1: As discussed, we decided to separate the auto-opening of the group menus in DPF out into a separate PR.

frontend/src/modules/_shared/DataProviderFramework/ViewportLayoutMenu.tsx
Anders2303: Unsure if you're referring to the check or just the commented code – the if statement is just a safety check in case a developer accidentally adds a menu item that **isn't** a `ViewLayout`.

I've removed the comment, atleast

frontend/src/modules/_shared/DataProviderFramework/ViewportLayoutMenu.tsx
Anders2303: Good point, I hadn't really considered radio items yet. Instead of over-complicating the composed menu, I instead made the menu compose the menu manually, as that seemed more clean. This change also included moving some styling out of the composed menu and into the custom parts, so those should be re-reviewed

rubenthoms (approved):
Looking good!

PR #1516, opened 2026-02-27
Utils - Hook for debounced listeners and optionally controlled values

Anders2303:
Extracted from #1075 

Adds to utility hooks that can be useful in lib-components:
* one for debouncing an on-change callback
* one for optionally making a components prop controlled/uncontrolled


rubenthoms (changes requested):
A couple of comments.

Issue #736, opened 2024-09-23
`fmu-dataio` data requests


anders-kiaer:
This is not a webviz issue per se, but a placeholder/container for data requests communicated to `fmu-dataio` project:

- [ ] Abstract away things: https://github.com/equinor/fmu-dataio/issues/668 (will make it easier for user to start using webviz modules, and easier maintenance on fmu-dataio side).
- [ ] Parent grid: https://github.com/equinor/fmu-dataio/issues/278
- [ ] Blocked/modelled well logs: https://github.com/equinor/fmu-dataio/issues/372
- [ ] FIPNUM vs region/zone mapping: https://github.com/equinor/fmu-dataio/issues/532
- [ ] Observations
- [ ] Well uuid / alias mapping
- [ ] Richer ensemble metadata https://github.com/equinor/fmu-dataio/issues/299

anders-kiaer:
**Actions from meeting Sep 23rd:**
- [ ] SUMO make `data.geometry` metadata searchable (introduced in https://github.com/equinor/fmu-dataio/pull/677).

PR #1559, opened 2026-03-17
Add FIP to region mapping

HansKallekleiv:
Relates to https://github.com/equinor/webviz/issues/736
- Queries an ensemble for fipnum standard result (simulator_fipregions_mapping) added in https://github.com/equinor/fmu-dataio/pull/1571
- Adds a small helper class to search the fipnum array, accessible from RegularEnsemble


frontend/tests/unit/EnsembleFipRegionsMapping.test.ts
```diff
@@ -0,0 +1,89 @@
...
+            "CentralSouth",
+            "CentralNorth",
+            "NorthHorst",
+            "CentralRamp",
+            "CentralHorst",
+            "EastLowland",
+        ]);
+
+        expect(mapping.getRegionsForZones(["Valysar", "Therys"]).length).toBe(7);
+        expect(mapping.getRegionsForZones([]).length).toBe(7);
```
Anders2303, line 77: If we wanna be extra pedantic with these, you could consider writing out the exact names you'd expect, but it's fine as is

Anders2303 (approved):
Looks good to me

PR #1563, opened 2026-03-19
[Tooling]: Derive modules list from found persistence files and detect violations of persistence file pattern

rubenthoms:


Anders2303 (commented):
Running this branch, the module map updated, re-adding the non-persistable modules that were removed earlier in this PR. Was this intentional?

Anders2303 (approved):

PR #1571, opened 2026-03-26
[DPF]: Fix shared setting intercept

jorgenherje:
Closes: #1569

Suggestions to resolve issue with Shared setting at root and in group. With same shared setting in group and at root, the constraints and values become incorrect.

### DIFF
--- a/frontend/src/modules/_shared/DataProviderFramework/framework/ExternalSettingController/ExternalSettingController.ts
+++ b/frontend/src/modules/_shared/DataProviderFramework/framework/ExternalSettingController/ExternalSettingController.ts
@@ -17,8 +17,8 @@ export class ExternalSettingController<
     TExternalValue extends SettingTypeDefinitions[TSetting]["externalValue"] | null =
         | SettingTypeDefinitions[TSetting]["externalValue"]
         | null,
-    TValueConstraints extends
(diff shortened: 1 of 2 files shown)


frontend/src/modules/_shared/DataProviderFramework/framework/ExternalSettingController/ExternalSettingController.ts
```diff
@@ -207,8 +210,10 @@ export class ExternalSettingController<
 
     makeIntersectionOfValueConstraints(): void {
         if (!this.syncStateFromControlledSettings()) {
-            // Not ready yet, but we should avoid exposing deprecated value constraints to the controlled settings
-            this._setting.setValueConstraints(null);
+            // Not ready yet — syncStateFromControlledSettings has already set loading=true on the
+            // output setting. Do NOT call setValueConstraints(null) here: doing so would
+            // force-initialize intermediate controlled settings with null, which cascades a fake
+            // "ready" signal back up the chain before async dependencies have resolved.
```
rubenthoms: You can shorten the comment as the second part ("Do NOT ...") is not required to be included here when the old code is gone.

rubenthoms (approved):

PR #1580, opened 2026-04-08
Prevent row number limit in `SimulationTimeSeries` to become `Nan`

anders-kiaer:
Closes #1579.


jorgenherje (changes requested):
See comments regarding using "correct" signal :)

PR #1630, opened 2026-04-22
Fix tooltip positioning

HansKallekleiv:
The deps update (#1586) bumped @equinor/eds-core-react from 0.48 to 2.4. The v2 Tooltip reads the anchor child's ref via children.props.ref (React 19 convention) and unconditionally overrides the child's ref through cloneElement. Since webviz is still on React 18 — where refs are not part of props — any ref forwarded into a Tooltip-wrapped child is silently dropped.

This broke DenseIconButton's forwarded ref, which caused ContextHelp's anchorRef.current to stay null and the popup to fall back to { top: 100, left: 100 } (upper-left of the viewport) instead of anchoring to the help button.

Fix: Wrap the internal button in DenseIconButton with a small inner forwardRef component that receives the outer ref via a plain prop (outerButtonRef) and merges it with the ref assigned by Tooltip.cloneElement. This bypasses the ref-replacement by routing the outer ref through props rather than the React ref channel.

### DIFF
--- a/frontend/src/lib/components/DenseIconButton/denseIconButton.tsx
+++ b/frontend/src/lib/components/DenseIconButton/denseIconButton.tsx
@@ -29,6 +29,38 @@ export type DenseIconButtonProps = {
     className?: string;
 };
 
+// NOTE: The wrapped EDS `Tooltip` clones its child and replaces the child's `ref`, reading the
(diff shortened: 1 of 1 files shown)


Anders2303 (approved):
Looks okay as a hotfix for now; I'll make an issue at the EDS repo and investigate a better fix later

PR #1652, opened 2026-05-06
Add sensitivity mean support to SensitivityPlot

### LINKED ISSUE #1640: Display mean for sensitivities
For sensitivities where each sensitivity only has seed variation and only one case for each sensitivity we should display the mean for the entire sensitivity (in addition to the p10 / p90 cases we currently calculate).
- As column in table
- Optional text box at zero line in plot?

HansKallekleiv:
Closes #1640 

Computes a mean response value for each sensitivity and exposes it in the sensitivity table. Active only for Monte Carlo sensitivities with the mean is calculated across all realizations in the single p10_p90 case. For scenario cases a mean across cases does not make much sense.

The chart now includes an optional Show mean points overlay. When enabled, one marker is shown per sensitivity at the scaled mean response value, making it easier to compare where each sensitivity average sits relative to the P10/P90.


anders-kiaer (approved):
LGTM 👍

PR #1654, opened 2026-05-08
[Bug]: Replace usage of `onChange` with `onValueChange` for numerical `Input`-component 

### LINKED ISSUE #1643: Crash when deleting steps in Color Scale Selector
Hi, I get an error when trying to delete the number of steps in the color selector. I want only two colors, and thought I could delete the '10' before typing '2' instead of having to click the arrow, but got 

Image

Image

when clicking the 'Report error' I got 

Image

:)

jorgenherje:
Numerical input components should not trigger too eagerly, `onChange` emits on every input change, which can create invalid intermediate states when editing numerical values, especially Nan values when clearing the input.

`onValueChange` fires when loosing focus, and ensures value within min/max.

---

Closes: #1579
Closes: #1643

### DIFF
--- a/frontend/src/framework/internal/components/RightSettingsPanel/private-components/colorPaletteSettings.tsx
+++ b/frontend/src/framework/internal/components/RightSettingsPanel/private-components/colorPaletteSettings.tsx
@@ -92,12 +92,7 @@ export const ColorPaletteSettings: React.FC<ColorPaletteSettingsProps> = (props)
                                 min={2}
                                 max={100}
                                 defaultValue={steps[ColorScaleDiscreteSteps.Sequential]}
-                                onChange={(e) =>
-                                    handleColorPaletteStepsChanged(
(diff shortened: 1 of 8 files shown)


frontend/src/framework/internal/components/RightSettingsPanel/private-components/colorPaletteSettings.tsx
rubenthoms, line 1: I think here we can keep the `onChange` as it is nice to get immediate feedback. No crashes occuring when clearing the input.

rubenthoms (approved):
One comment. Otherwise, LGTM!

frontend/src/framework/internal/components/RightSettingsPanel/private-components/colorPaletteSettings.tsx
jorgenherje: As discussed: We'll leave as it is

PR #1660, opened 2026-05-12
[Tooling]: Added caching for Playwright (tests), and backend docker container when checking auto-generated API code

rubenthoms:
### Summary

- Cache Playwright browser binaries across runs, keyed on `frontend/package-lock.json`. On a cache hit, only OS-level system dependencies are installed (`install-deps`); the ~300 MB browser download is skipped.
- Cache Docker build layers for the backend image used in the API sync check, using Docker Buildx's native GitHub Actions cache backend (`type=gha,mode=max`). Unchanged layers (e.g. the Poetry dependency install) are reused across runs; only layers whose inputs have changed are rebuilt.

### Cache invalidation

**Playwright** — the cache key includes a hash of `package-lock.json`, so it is automatically busted whenever the Playwright version is bumped.

**Docker** — invalidation is content-addressed at the layer level by Docker itself. The `COPY ./backend_py/primary` instruction in the Dockerfile means any change to backend source files (routes, schemas, response models) will bust that layer and all subsequent ones, ensuring the API sync check always runs against the correct backend image.

### Test plan
- [x] Verify the Playwright cache is populated on the first run and restored on subsequent runs with no Playwright version change
- [x] Verify the Docker layer cache is populated on the first run and partially restored (deps layer hit, source layer miss) after a backend source-only change
- [x] Verify the API sync check correctly fails when the backend API changes but the generated frontend code has not been updated


.github/workflows/webviz.yml
```diff
@@ -73,14 +87,55 @@ jobs:
...
+          set -euo pipefail
+
+          # Build using the local BuildKit cache for layer-level caching.
+          # Writes to a separate directory to avoid corrupting the restored cache
+          # mid-build; the swap below atomically replaces it for the next run.
+          docker buildx build \
+            -f ./backend_py/primary/Dockerfile \
+            -t backend:latest \
+            --cache-from type=local,src=/tmp/buildx-cache \
+            --cache-to type=local,dest=/tmp/buildx-cache-new,mode=max \
```
anders-kiaer, line 116: I see we tried the `buildx` cache backend built for GitHub actions in an earlier commit (`type=gha`). What was the reason we wanted to change to local? Should we add a comment here stating the reason?

.github/workflows/webviz.yml
```diff
@@ -73,14 +87,55 @@ jobs:
...
+          key: backend-docker-${{ runner.os }}-${{ hashFiles('backend_py/primary/Dockerfile', 'backend_py/primary/pyproject.toml', 'backend_py/primary/poetry.lock', 'backend_py/libs/**') }}
+          restore-keys: |
+            backend-docker-${{ runner.os }}-
+
       - name: 🕵️ Check auto-generated frontend code is in sync with backend
         run: |
-          docker build -f ./backend_py/primary/Dockerfile -t backend:latest .
-          CONTAINER_ID=$(docker run --detach -p 5000:5000 --env UVICORN_PORT=5000 --env WEBVIZ_SKIP_LIFESPAN_GENERATE_API_ONLY=true --env WEBVIZ_CLIENT_SECRET=0 --env WEBVIZ_SMDA_SUBSCRIPTION_KEY=0 --env WEBVIZ_VDS_HOST_ADDRESS=0 --env WEBVIZ_ENTERPRISE_SUBSCRIPTION_KEY=0 backend:latest)
-          sleep 10  # Ensure the backend server is up and running exposing /openapi.json
+          set -euo pipefail
```
anders-kiaer, line 107: Is this needed? [`-e` seems default on GitHub actions](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsshell), and we do not use pipes in this job?

.github/workflows/webviz.yml
```diff
@@ -73,14 +87,55 @@ jobs:
...
+            --env WEBVIZ_CLIENT_SECRET=0 \
+            --env WEBVIZ_SMDA_SUBSCRIPTION_KEY=0 \
+            --env WEBVIZ_VDS_HOST_ADDRESS=0 \
+            --env WEBVIZ_ENTERPRISE_SUBSCRIPTION_KEY=0 \
+            backend:latest)
+
+          # Single-quoted so $CONTAINER_ID is expanded at exit time, not now.
+          trap 'docker stop "$CONTAINER_ID"' EXIT
+
+          sleep 10
```
anders-kiaer, line 136: Replacing this by something more explicit which also continues immediately when server is up?
```suggestion
          # Waiting on OpenAPI server up and running:
          timeout 10 bash -c 'until curl -sf http://localhost:5000/openapi.json > /dev/null 2>&1; do sleep 1; done'
```

.github/workflows/webviz.yml
```diff
@@ -73,14 +87,55 @@ jobs:
...
+            --env UVICORN_PORT=5000 \
+            --env WEBVIZ_SKIP_LIFESPAN_GENERATE_API_ONLY=true \
+            --env WEBVIZ_CLIENT_SECRET=0 \
+            --env WEBVIZ_SMDA_SUBSCRIPTION_KEY=0 \
+            --env WEBVIZ_VDS_HOST_ADDRESS=0 \
+            --env WEBVIZ_ENTERPRISE_SUBSCRIPTION_KEY=0 \
+            backend:latest)
+
+          # Single-quoted so $CONTAINER_ID is expanded at exit time, not now.
+          trap 'docker stop "$CONTAINER_ID"' EXIT
```
anders-kiaer, line 134: Good and more robust replacement of `docker stop $CONTAINER_ID`. We could also consider removing this completely as GitHub actions jobs are independent it should/will stop automatically when the job ends.

anders-kiaer (commented):
Looks good - would be nice to improve `frontend` CI speed 👍 

Some questions/comments below, otherwise LGTM.

.github/workflows/webviz.yml
rubenthoms: You are right that `-e` is enabled by default and that we don't have any pipes. However, adding `-u` adds immediate value by adding fail-fast behaviour. `-o pipefail` is an insurance for future changes to the step as it discovers any hidden failures (e.g. `some_command | grep ...` where grep is successful but some_command fails). So yes, theoretically, we could only add `-u` but `set -euo pipefail` is a good default and immediately shows what is set (without having to know about and/or rely on GitHub).

PR #1681, opened 2026-06-02
Prepare for using passwords stored in Azure KeyVault with Redis

sigurdp:
Configures `radixconfig.yml` so that secrets can be consumed by `redis-auth-store` and `redis-cache` components
Also renames the `redis-user-session` to `redis-auth-store`

Done so that we can configure workload identity in Radix/Azure before activating code that relies on the passwords.


anders-kiaer (approved):
LGTM 👍

PR #1684, opened 2026-06-04
Fix: Replace fixed sleep with curl retry to reliably wait for backend in API generation CI check

rubenthoms:
The "Check auto-generated frontend code is in sync with backend" CI step was intermittently failing with an HTTP error from `npm run generate:api`. A possible cause might have been a race condition: the backend startup includes a synchronous outbound HTTP call at import time (`config.py`) whose latency varies in CI and based on SUMO load, meaning the fixed `sleep 10` was sometimes not enough.

- Replace the fixed `sleep 10` with `curl --retry` to poll `/alive` until the server is actually ready, rather than assuming it's up after an arbitrary delay
- Keep the initial `sleep 10` as a baseline to avoid hammering the port before the container process has started at all
- Add `docker logs` after the generate step so backend errors are visible in CI output when the step fails


sigurdp (approved):
👍

PR #1720, opened 2026-06-12
Time metadata per property for grid properties

HansKallekleiv:
- Fetch t0,t1 from Sumo for each grid property, instead of the union.
- Position time dropdown correctly in 3d grid layer


sigurdp (approved):
👍  PR looks good.
As discussed in stand-up it does uncover some shortcomings in the client UI, but they will have to be fixed seoarately.

PR #1721, opened 2026-06-12
[EDS] Migrated Inplace Volumetrics modules

Anders2303:
Resolves:
- #1696 
- #1697 
- #1698


HansKallekleiv (changes requested):
Looks good, but there are issues with table sorting.
`InplaceVolumesNew` should have table sorting (might be that it missing currently in prod, but it should )
`InplaceVolumesTable` seems to have sorting enabled, but no sorting is performed when clicking in a column

HansKallekleiv (approved):

PR #1723, opened 2026-06-15
Rft from standard results with eds

HansKallekleiv:
Reworks the RFT module to use standard-results , including observations.

### Backend
- New endpoint for observations
- Minor changes in rft access layer

### Frontend
- Refactor RFT similar to RelPerm
- Added datachannel with draggable depth slider with instant update
- Fixed a resizing issue (also added to RelPerm)

Closes #1685


frontend/src/modules/RelPerm/view/view.tsx
rubenthoms, line 1: Is this change to a different module intentional in this PR?

frontend/src/modules/Rft/channelDefs.ts
```diff
@@ -0,0 +1,14 @@
+import type { ChannelDefinition } from "@framework/types/dataChannnel";
+import { KeyKind } from "@framework/types/dataChannnel";
+
+export enum ChannelIds {
+    RESPONSE_AT_DEPTH = "Response value (per realization at selected depth)",
```
rubenthoms: String enum value representations should be the same as their respective enum value:
`RESPONSE_AT_DEPTH = "RESPONSE_AT_DEPTH"`

frontend/src/modules/Rft/settings/atoms/derivedAtoms.ts
```diff
@@ -1,71 +1,127 @@
...
-    const availableRftTimestampsUtcMs = get(availableRftTimestampsUtcMsAtom);
-    const userSelectedRftTimestampsUtcMs = get(userSelectedRftTimestampsUtcMsAtom);
-    return fixupSelectedOrFirstValue(userSelectedRftTimestampsUtcMs, availableRftTimestampsUtcMs);
+export const selectedTimestampUtcMsAtom = atom<number | null>((get) => {
+    return fixupSelectedOrFirstValue(get(userSelectedTimestampUtcMsAtom).value, get(availableTimestampsUtcMsAtom));
 });
+
+export const rftDataAccessorStatusAtom = atom((get) => get(rftRealizationDataQueriesAtom));
+
+export const rftObservationsStatusAtom = atom((get) => get(rftObservationsQueriesAtom));
```
rubenthoms: What are these atoms for as they are only copying the query-atoms? In addition, the names are very misleading as they are used to actually access data in the view. Hence, they are holding way more than just the status - consider changing the name to reflect their actual nature. If you remove these copies, consider changing the interface as well. A proper name could be "*Result*.

frontend/src/modules/Rft/settings/atoms/queryAtoms.ts
```diff
@@ -1,54 +1,198 @@
...
+                query: {
+                    case_uuid: ensembleIdent.getCaseUuid(),
+                    ensemble_name: ensembleIdent.getEnsembleName(),
+                    well_name: selectedWellName ?? "",
+                    response_name: selectedResponseName ?? "",
+                    timestamps_utc_ms: selectedTimestampUtcMs !== null ? [selectedTimestampUtcMs] : null,
+                    realizations_encoded_as_uint_list_str: realizationsEncodedAsUintListStr,
+                    ...makeCacheBustingQueryParam(ensembleIdent),
+                },
+            });
```
rubenthoms: What is the reason for separating the functions out? I think this makes it harder to read as the viewer has to jump to the respective function first.

frontend/src/modules/Rft/settings/atoms/queryAtoms.ts
```diff
@@ -1,54 +1,198 @@
...
+                    })
+                    .filter(function isNonNullError(error): error is RftApiError {
+                        return error !== null;
+                    }),
+            };
+        },
     };
-    return query;
-});
+}
```
rubenthoms: Same here.

frontend/src/modules/Rft/view/view.tsx
```diff
@@ -0,0 +1,160 @@
...
+    const rftObservationsStatus = viewContext.useSettingsToViewInterfaceValue("rftObservationsStatus");
+
+    const [dataChannelDepth, setDataChannelDepth] = useAtom(dataChannelDepthAtom);
+    const depthLineOverlay = usePlotOverlay();
+
+    const propagatedErrorMessages = usePropagateAllApiErrorsToStatusWriter(rftDataAccessorStatus.errors, statusWriter);
+    const propagatedErrorMessage = propagatedErrorMessages[0] ?? null;
+    const instanceTitle = makeRftPlotTitle(wellName, responseName, timestampUtcMs);
+
+    statusWriter.setLoading(rftDataAccessorStatus.isFetching);
```
rubenthoms: Why isn't settings owning its own status propagation? This is a bit of a design tension to me. Currently, both status and data are passed to the view just to report potential errors and loading states in the view while settings could actually handle that itself. View is only interested in the data. Consider just sending the data and maybe an additional isFetching status just for the view to display a spinner (if you want that). Anyways, the module is in any case showing a loading bar at the top.

frontend/src/modules/Rft/view/utils/plotData.ts
```diff
@@ -0,0 +1,69 @@
...
+    const dateString = timestampUtcMsToCompactIsoString(timestampUtcMs).split("T")[0];
+
+    for (const ensembleObservations of observationsData) {
+        const matchingGroup = ensembleObservations.observations.find(
+            (group) => group.well_name === wellName && group.date.split("T")[0] === dateString,
+        );
+        if (matchingGroup) {
+            return matchingGroup.observations.filter(
+                (observation) => observation.property.toUpperCase() === responseName.toUpperCase(),
+            );
```
rubenthoms: Is it by design that you only extract observations for the first ensemble with a match? If yes, consider reflecting this in the function name.

frontend/src/modules/Rft/view/view.tsx
```diff
@@ -0,0 +1,160 @@
...
+    const dataAccessor = rftDataAccessorStatus.dataAccessor;
+    const entries = React.useMemo(() => dataAccessor?.getEntries() ?? [], [dataAccessor]);
+    const selectedEnsembles = React.useMemo(
+        () =>
+            entries.flatMap((entry) => {
+                const ensemble = ensembleSet.findEnsemble(entry.ensembleIdent);
+                return ensemble ? [ensemble] : [];
+            }),
+        [entries, ensembleSet],
+    );
```
rubenthoms, line 61: This loop produces a lot of duplicates as there is no deduping.

frontend/src/modules/Rft/view/utils/RftPlotBuilder.ts
```diff
@@ -0,0 +1,423 @@
...
+    return Array.from(depthSet).sort((left, right) => left - right);
+}
+
+function sortCurveByDepth(depths: number[], values: number[]): { depths: number[]; values: number[] } {
+    const orderedIndices = depths.map((_, index) => index).sort((left, right) => depths[left] - depths[right]);
+    return {
+        depths: orderedIndices.map((index) => depths[index]),
+        values: orderedIndices.map((index) => values[index]),
+    };
+}
```
rubenthoms: Isn't there already a function for this in `utils/RftDataAccessor`? Consider moving this to a separate util file.

frontend/src/modules/Rft/view/utils/RftPlotBuilder.ts
```diff
@@ -0,0 +1,423 @@
...
+
+export function interpolateCurveValueAtDepth(depths: number[], values: number[], targetDepth: number): number {
+    const sorted = sortCurveByDepth(depths, values);
```
rubenthoms: This is only called by the data generator which already works with pre-sorted curves (all come through the `RftDataAccessor`). So this looks like a redundant call. Consider to either remove it or document why it is necessary.

frontend/src/modules/Rft/settings/settings.tsx
```diff
@@ -0,0 +1,330 @@
...
+                </SettingWrapper.Section>
+                <SettingWrapper.Section title="Data channel">
+                    <SettingWrapper label="Depth line">
+                        <CheckboxCompositions.WithLabel
+                            label="Show depth line in plot"
+                            checked={showDepthLine}
+                            onCheckedChange={setShowDepthLine}
+                            size="small"
+                        />
+                    </SettingWrapper>
```
rubenthoms: Consider using a SettingWrapper without a label and use a `SwitchCompositions.WithLabel`.

frontend/src/modules/Rft/settings/settings.tsx
```diff
@@ -0,0 +1,330 @@
...
+                        loadingOverlay={tableDefinitionsArePending}
+                        errorOverlay={tableDefinitionsErrorMessage}
+                        stacked
+                    >
+                        <Select
+                            options={makeTimestampOptions(availableTimestampsUtcMs)}
+                            value={selectedTimestampUtcMs !== null ? [selectedTimestampUtcMs.toString()] : []}
+                            onValueChange={handleTimestampChange}
+                            filter={availableTimestampsUtcMs.length > 6}
+                            size={Math.max(1, Math.min(availableTimestampsUtcMs.length, 6))}
```
rubenthoms: I find these adjusting size selects a bit odd - I would at least vote for having a min size of 3 items.

rubenthoms (changes requested):
In the app, the module looks nice and is easy to use :+1: However, there are some implementation details that require changes/fixes before we can take this in. I stopped further review for now as these changes might refactor parts of the code.

frontend/src/modules/Rft/settings/atoms/derivedAtoms.ts
```diff
@@ -1,71 +1,123 @@
...
+    return {
+        showIndividualRealizations: get(showIndividualRealizationsAtom),
+        showStatisticalLines: get(showStatisticalLinesAtom),
+        showStatisticalFan: get(showStatisticalFanAtom),
+        showObservations: get(showObservationsAtom),
+        selectedStatistics: get(selectedStatisticsAtom),
+    };
 });
+
+export const showDepthLineSettingAtom = atom((get) => get(showDepthLineAtom));
```
rubenthoms: Seems unnecessary?

frontend/src/modules/Rft/settings/atoms/derivedAtoms.ts
```diff
@@ -1,71 +1,123 @@
...
-    const userSelectedRftTimestampsUtcMs = get(userSelectedRftTimestampsUtcMsAtom);
-    return fixupSelectedOrFirstValue(userSelectedRftTimestampsUtcMs, availableRftTimestampsUtcMs);
+export const visualizationSettingsAtom = atom((get) => {
+    return {
+        showIndividualRealizations: get(showIndividualRealizationsAtom),
+        showStatisticalLines: get(showStatisticalLinesAtom),
+        showStatisticalFan: get(showStatisticalFanAtom),
+        showObservations: get(showObservationsAtom),
+        selectedStatistics: get(selectedStatisticsAtom),
+    };
```
rubenthoms: This seems to be the wrong place for combining the settings. Seems to be only there for the interface - why not doing this directly in the interface then?

frontend/src/modules/Rft/settings/atoms/derivedAtoms.ts
```diff
@@ -1,71 +1,69 @@
...
-        return values[0];
+function intersectArrays<T>(arrays: T[][]): T[] {
+    if (arrays.length === 0) {
+        return [];
     }
-    return null;
+
+    return arrays.reduce<T[]>((intersection, currentArray) => {
+        return intersection.filter((value) => currentArray.includes(value));
+    }, arrays[0]);
```
rubenthoms: Minor: You could consider to exclude the first array element in order to avoid having it being compared with itself.
```return arrays.slice(1).reduce<T[]>((intersection, currentArray) => {
    return intersection.filter((value) => currentArray.includes(value));
}, arrays[0]);```

More performant also with  a set:
```const set = new Set(currentArray);
return intersection.filter((value) => set.has(value));```

File: frontend/src/modules/Rft/settings/atoms/derivedAtoms.ts [hunk 15]
```diff
@@ -1,71 +1,69 @@
...
-});
+    return tableDefinitionQueries.flatMap((queryResult, index) => {
+        if (!queryResult.data) {
+            return [];
+        }
 
-export const availableRftResponseNamesAtom = atom<string[]>((get) => {
-    const rftTableDefinition = get(rftTableDefinitionAtom);
-    return rftTableDefinition.data?.response_names.map((item) => item) ?? [];
+        return [{ ensembleIdent: selectedEnsembleIdents[index], tableDefinition: queryResult.data }];
```
rubenthoms: Consider safe-guarding for `selectedEnsembleIdents[index] === undefined`

```const ensembleIdent = selectedEnsembleIdents[index];
if (!ensembleIdent) return [];
return [{ ensembleIdent, tableDefinition: queryResult.data }];```

File: frontend/src/modules/Rft/settings/atoms/persistableFixableAtoms.ts [hunk 16]
```diff
@@ -0,0 +1,81 @@
...
+    },
+    fixupFunction: ({ value, precomputedValue: availableWellNames }) => {
+        const fixedSelection = fixupUserSelection([value], availableWellNames);
+        return fixedSelection?.[0] ?? null;
+    },
+});
+
+export const selectedTimestampUtcMsAtom = persistableFixableAtom<number | null, number[]>({
+    initialValue: null,
+    areEqualFunction: isEqual,
```
rubenthoms: This is an overkill for a simple number comparison - consider removing this line.

frontend/src/modules/Rft/settings/persistence.ts
```diff
@@ -0,0 +1,90 @@
...
+    dataChannelDepth: get(dataChannelDepthAtom),
+    showDepthLine: get(showDepthLineAtom),
+});
+
+export const deserializeSettings: DeserializeStateFunction<SerializedSettings> = (raw, set) => {
+    setIfDefined(
+        set,
+        selectedEnsembleIdentsAtom,
+        raw.selectedEnsembleIdentStrings?.map((ensembleIdentString) =>
+            RegularEnsembleIdent.fromString(ensembleIdentString),
```
rubenthoms: Missing error handling here.

frontend/src/modules/Rft/settings/atoms/queryAtoms.ts
```diff
@@ -1,54 +1,157 @@
...
+                ...makeCacheBustingQueryParam(ensembleIdent),
             },
-        }),
-        enabled: Boolean(selectedEnsembleIdent?.getCaseUuid() && selectedEnsembleIdent?.getEnsembleName()),
-    };
-    return query;
+        });
+
+        return function getRftTableDefinitionQueryOptions() {
+            return { ...options, enabled: Boolean(ensembleIdent) };
```
rubenthoms: `Boolean(ensembleIdent)` is always true?

frontend/src/modules/Rft/settings/atoms/queryAtoms.ts
```diff
@@ -1,54 +1,157 @@
...
+        return {
+            observationsData,
+            errors: results
+                .map(function getResultError(result) {
+                    return result.error;
+                })
+                .filter(function isNonNullError(error): error is RftApiError {
+                    return error !== null;
+                }),
+        };
```
rubenthoms, line 158: Should we add `isFetching` and `isError` to the combined result?

frontend/src/modules/Rft/utils/RftDataAccessor.ts
```diff
@@ -0,0 +1,44 @@
...
+        });
+    }
+
+    getEntries(): RftRealizationCurve[] {
+        return this._entries;
+    }
+
+    getEnsembleIdents(): RegularEnsembleIdent[] {
+        return this._ensembleIdents;
+    }
```
rubenthoms: Consider return as `readonly` to signal that these arrays should not be mutated.

frontend/src/modules/Rft/utils/curveUtils.ts
```diff
@@ -0,0 +1,30 @@
+export function sortCurveByDepth(depths: number[], values: number[]): { depths: number[]; values: number[] } {
+    const orderedIndices = depths.map((_, index) => index).sort((left, right) => depths[left] - depths[right]);
```
rubenthoms, line 7: Consider adding a check for equal length of depths and values.

frontend/src/modules/Rft/utils/curveUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+        const upperDepth = sortedDepths[index + 1];
+        if (targetDepth >= lowerDepth && targetDepth <= upperDepth) {
+            if (upperDepth === lowerDepth) {
+                return sortedValues[index];
+            }
+            const fraction = (targetDepth - lowerDepth) / (upperDepth - lowerDepth);
+            return sortedValues[index] + fraction * (sortedValues[index + 1] - sortedValues[index]);
+        }
+    }
+
```
rubenthoms, line 33: Consider adding a comment: `// Single-element array: targetDepth must equal sortedDepths[0], return its value` or something similar

frontend/src/modules/Rft/view/components/DepthLineOverlay.tsx
```diff
@@ -0,0 +1,159 @@
...
+            style={{ top: geometry.top, left: geometry.left, width: geometry.width }}
+            onPointerDown={handlePointerDown}
+            onPointerMove={handlePointerMove}
+        >
+            <div className="bg-surface/70 py-4xs px-xs absolute -top-2 right-0 min-h-4 text-right bg-blend-overlay">
+                {depthLabel}
+            </div>
+            <div
+                className={resolveClassNames(
+                    "bg-neutral-strong group-hover:bg-accent-strong-hover group-active:bg-accent-strong-active h-0.5 w-full",
```
rubenthoms: Compared to your previous implementation, this one is not using a lighter border in order to work on darker backgrounds. You can consider to add an outline top and bottom with a lighter color:
`outline-neutral-subtle outline`

frontend/src/modules/Rft/view/utils/RftPlotBuilder.ts
```diff
@@ -0,0 +1,420 @@
...
+
+        for (const entry of this._dataAccessor.getEntries()) {
+            const ensembleKey = entry.ensembleIdent.toString();
+            if (!firstEntryByEnsemble.has(ensembleKey)) {
+                firstEntryByEnsemble.set(ensembleKey, entry);
+            }
+        }
+
+        return Array.from(firstEntryByEnsemble.entries()).map(([ensembleKey, entry]) => {
+            shownLegendEnsembles.add(ensembleKey);
```
rubenthoms, line 61: This mutation is a non-obvious side-effect and makes the function impure. Returning only the set of ensembles created during this function call is a cleaner approach: `{ traces: Partial<PlotData>[], shownLegendEnsembles: Set<string> }`

frontend/src/modules/Rft/view/utils/RftPlotBuilder.ts
```diff
@@ -0,0 +1,420 @@
...
+        }
+
+        return Array.from(groups.values());
+    }
+
+    private makeEnsembleColor(ensembleIdent: RegularEnsembleIdent): string {
+        const ensemble = this._selectedEnsembles.find(
+            (candidate) => candidate.getIdent().toString() === ensembleIdent.toString(),
+        );
+        return ensemble?.getColor() ?? this._colorSet.getFirstColor();
```
rubenthoms: This fallback could be very misleading as missing ensembles would get the same color as ensemble no 1. Consider add a console.error and return a neutral color (some gray) here.

frontend/src/modules/Rft/view/utils/RftPlotBuilder.ts
```diff
@@ -0,0 +1,420 @@
...
+}
+
+function resampleEntriesToCommonDepths(entries: RftRealizationCurve[]): RftRealizationCurve[] {
+    const commonDepths = makeCommonDepthGrid(entries);
+    if (commonDepths.length === 0) {
+        return [];
+    }
+
+    return entries.map((entry) => {
+        const sorted = sortCurveByDepth(entry.depths, entry.values);
```
rubenthoms: This does also seem to be a redundant call as all entries are coming from RftDataAccessor which is already sorting them by depth?

frontend/src/modules/Rft/view/utils/RftPlotBuilder.ts
```diff
@@ -0,0 +1,420 @@
...
+
+function makeTransparentColor(color: string, opacity: number): string {
+    if (!color.startsWith("#") || color.length !== 7) {
+        return color;
+    }
+
+    const red = parseInt(color.slice(1, 3), 16);
+    const green = parseInt(color.slice(3, 5), 16);
+    const blue = parseInt(color.slice(5, 7), 16);
+    return `rgba(${red}, ${green}, ${blue}, ${opacity})`;
```
rubenthoms: Consider using `culori` for this task:
```
const parsed = parse(color);
if (!parsed) return color;
return formatRgb({ ...parsed, alpha: opacity });
```

frontend/src/modules/Rft/view/hooks/useRftPlotBuilder.ts
```diff
@@ -0,0 +1,105 @@
...
+                    ? plotBuilder.makeIndividualRealizationTraces(responseName, shownLegendEnsembles)
+                    : []),
+                ...(visualizationSettings.showObservations
+                    ? plotBuilder.makeObservationTraces(observationsPerEnsemble, responseName, shownLegendEnsembles)
+                    : []),
+            ];
+
+            const layout = plotBuilder.makeLayout(size, responseName, valueRange);
+
+            return { plotData, layout };
```
rubenthoms: Consider splitting this up as every resize would trigger both layout and data recomputation. Two memos, one for layout and one for plotData would be more performant.

frontend/src/modules/Rft/settings/settings.tsx
```diff
@@ -0,0 +1,335 @@
...
+
+    const validRealizations = React.useMemo(() => {
+        const realizationSet = new Set<number>();
+        for (const ensembleIdent of selectedEnsembleIdents) {
+            for (const realization of filterEnsembleRealizationsFunc(ensembleIdent)) {
+                realizationSet.add(realization);
+            }
+        }
+        return Array.from(realizationSet).sort((a, b) => a - b);
+    }, [filterEnsembleRealizationsFunc, selectedEnsembleIdents]);
```
rubenthoms: Consider a derived atom instead to avoid re-renders here. Also cleaner regarding business logic separation.

frontend/src/modules/Rft/settings/settings.tsx
```diff
@@ -0,0 +1,335 @@
...
+                        />
+                    </SettingWrapper>
+                    <SettingWrapper
+                        label="Date"
+                        annotations={timestampAnnotations}
+                        loadingOverlay={tableDefinitionsArePending}
+                        errorOverlay={tableDefinitionsErrorMessage}
+                        stacked
+                    >
+                        <Select
```
rubenthoms: Consider using a `<Select<number> />` here to avoid having to transform from string to number and vice-versa.

frontend/src/modules/Rft/settings/settings.tsx
```diff
@@ -0,0 +1,335 @@
...
+                        />
+                    </SettingWrapper>
+                </SettingWrapper.Section>
+            </SettingWrapper.Group>
+        </Collapsible.ScrollArea>
+    );
+}
+
+function arraysHaveSameValues(left: string[], right: string[]): boolean {
+    return left.length === right.length && left.every((value) => right.includes(value));
```
rubenthoms: Consider using lodash's `xor` util function for this: `return xor(left, right).length === 0;`

rubenthoms (changes requested):
Nice improvements! There are still a couple of issues and considerations.

frontend/src/modules/Rft/settings/atoms/derivedAtoms.ts
HansKallekleiv: these are aligned by construction, so isnt this an impossible sitatuation? Better to keep the code lean?

rubenthoms: Might be safer regarding a future refactor but you're right that under current conditions this scenario is actually impossible due to how Jotai is working. We can add it or not add it.

frontend/src/modules/Rft/settings/persistence.ts
rubenthoms:
```
setIfDefined(
    set,
    selectedEnsembleIdentsAtom,
    raw.selectedEnsembleIdentStrings?.flatMap((s) => {
        try {
            return [RegularEnsembleIdent.fromString(s)];
        } catch {
            return [];
        }
    }),
);
```

rubenthoms (approved):
I think we can this in now.

PR #1727, opened 2026-06-15
[EDS] Distribution plot

HansKallekleiv:
Closes #1708

### DIFF
--- a/frontend/src/modules/DistributionPlot/settings/settings.tsx
+++ b/frontend/src/modules/DistributionPlot/settings/settings.tsx
@@ -1,15 +1,14 @@
-import type React from "react";
-
 import { useAtom } from "jotai";
 
 import { useApplyInitialSettingsToState } from "@framework/InitialSettings";
 import type { ModuleSettingsProps } from "@framework/Module";
-import { Checkbox } from "@lib/components/Checkbox";
-import { CollapsibleGroup } from "@lib/components/CollapsibleGroup";
-import { Dropdown } from "@lib/components/Dropdown";
-import { Label } from "@lib/components/Label";
-import { RadioGroup } from "@lib/components/RadioGroup";
-import { Slider } from "@lib/components/Slider";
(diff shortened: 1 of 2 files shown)


frontend/src/modules/DistributionPlot/settings/settings.tsx
```diff
@@ -55,115 +54,92 @@ export function Settings({ initialSettings }: ModuleSettingsProps<Interfaces>) {
...
-            <CollapsibleGroup title="Plot type" expanded>
-                <Dropdown value={plotType as string} options={plotTypes} onChange={handlePlotTypeChanged} />
-            </CollapsibleGroup>
-            {makeContent()}
-        </div>
+        <Collapsible.ScrollArea>
+            <SettingWrapper.Group>
+                <SettingWrapper.Section title="Plot type" defaultOpen>
+                    <SettingWrapper label="Plot type">
+                        <Combobox items={plotTypes} value={plotType} onValueChange={handlePlotTypeChanged} />
```
jorgenherje: Combobox now has value `plotType` which is type `PlotType` enum. `onValueChange` returns `PlotType | null`, and we could consider one-liner `onValueChange={v => v && setPlotType(v)}` and remove `handlePlotTypeChanged` (it's not a callback anyways).

`plotTypes` could be renamed to `plotTypeItems` for clarity

jorgenherje (changes requested):
One minor comment regarding settings.

Also replace old tailwind `mb-2` with new - e.g. `mb-2xs` or something suitable - in view.tsx:

```
function MaxNumberPlotsExceededMessage(): React.ReactElement {
    return (
        <ContentWarning>
            <Warning fontSize="large" className="mb-2" />
            Too many plots to display. Due to performance limitations, the number of plots is limited to {MAX_NUM_PLOTS}
            .
        </ContentWarning>
    );
}
```

jorgenherje (approved):

PR #1737, opened 2026-06-16
[Deps] Replaced lodash dependency with lodash-es

Anders2303:
~`realizationFilterUtils` had an incorrect import of `lodash-es` which is not one of our dependencies.~

~A PR forever ago changed the type dependency to be `@types/lodash-es`, but there doesn't seem to be a reason to do this, at-least not anymore. Changed back to `@types/lodash` to avoid confusion~

As [discussed below](https://github.com/equinor/webviz/pull/1737#discussion_r3426713044) we've decided to swap the `lodash` dependency with `lodash-es`, which will offer better tree-shaking


frontend/package.json
```diff
@@ -92,7 +92,7 @@
         "@tailwindcss/vite": "^4.2.2",
         "@types/culori": "^4.0.0",
         "@types/geojson": "^7946.0.16",
-        "@types/lodash-es": "^4.17.9",
+        "@types/lodash": "^4.17.24",
```
rubenthoms: We should probably switch to `lodash-es` in order to improve tree-shaking and reduce bundle size.

rubenthoms (approved):
Adjust the imports mentioned by Copilot and the PR title and description. Otherwise, LGTM.

PR #1743, opened 2026-06-19
Create tutorial videos from integration tests

anders-kiaer:
- [X] Extend Playwright integration tests to fetching actual test data from Sumo.
- [X] Add a script that creates a test cookie in Playwright in order to bypass sign in page (such that we can test the app).
- [X] Add possibility to record video from the tests (`RECORD=1`). Add cursor marker as well as utility for narrating and slowing down (pacing) the video at specified steps (if not recording these are no-op calls).
- [X] Use `ffmpeg` to combine video and audio, and upload to Azure blob container on success.

This PR adds the test utilities for doing the recording + narration + uploading. Follow-up PR will extend the integration tests with more tests / stories (which then again can be used both as extended test base, and tutorial videos).


.github/workflows/e2e_recordings.yml
```diff
@@ -0,0 +1,114 @@
+name: e2e recordings
+
+on:
+  push:
+    # branches:
+    #  - main
```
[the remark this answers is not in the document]
anders-kiaer: Yes, this will be done before merge. Waiting until after review in case of changes (and in order to see the result from those changes).

frontend/tests/e2e/stories/simulationTimeSeriesModule.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+        // Make sure the active module's settings panel is expanded
+        const expandSettingsButton = page.getByTitle("Expand settings panel");
+        if (await expandSettingsButton.isVisible()) {
+            await smoothClick(page, expandSettingsButton);
+            await pace(page);
+        }
+
+        // Select a vector so the chart has something to plot
+        const vectorSelectorContainer = page.locator("div.cursor-text.min-w-48").first();
```
HansKallekleiv: Should we use a data-testid here?

HansKallekleiv (approved):

frontend/tests/e2e/stories/simulationTimeSeriesModule.test.ts
anders-kiaer: Good idea - added `data-testid`.

PR #1811, opened 2026-07-03
fix: 2D/3D viewer readout and improved number formatting utilities

rubenthoms:


frontend/src/modules/InplaceVolumesNew/view/utils/TableBuilder.ts
```diff
@@ -45,13 +45,13 @@ export function buildStatisticsTableData(groupedData: GroupedTableData, resultNa
             subplotValue: entry.subplotLabel,
             colorByValue: entry.colorLabel,
             colorByKey: entry.colorKey,
-            mean: formatNumber(stats.mean, 2),
```
HansKallekleiv, line 48: These should use SI system.

Image

Image

HansKallekleiv (approved):
Looks good, except one small thing that should be fixed. Approved

PR #1815, opened 2026-07-06
fix: more robust logic in order to be independent of call-order

rubenthoms:


Anders2303 (approved):
Seems to have solved the crash-issue I was having

PR #1819, opened 2026-07-09
[Lib, EDS] Slider improvements and GridLayerRangeSettings tweak

Anders2303:
* Split slider component code into multiple files
* Removed "always" display option for thumb labels
  * _Making the thumb always be visible caused issues with TooltipBase when hiding the anchor and resizing; causing notable missalignment issues. We'll need to re-implement the slider labels to something less error prone_
* Fixed some styling issues in slider
* Made slider thumb labels stay within width of track
* Allowed showing thumb label below the track
* Reverted GridLayerRangeSettings design to the old one; no locks, and numbers on each side of slider
  * I kept the min/max logic in the code, so it'd be easy to re-introduce it later
  * Also did some small restructuring to avoid duplicate code when rendering the "k" value row 
  * _Slider with an adjacent number field is used a lot, could consider making a composisiton for that_

image

(if setting panel grows to small, inputs are hidden)
image


frontend/src/lib/components/Slider/_components/markerLabel.tsx
```diff
@@ -0,0 +1,71 @@
...
+    value: number;
+    index: number;
+    numMarkers: number;
+    disabled?: boolean;
+    labelFormat?: (markerValue: number, index: number) => React.ReactNode;
+    onClick: (value: number) => void;
+};
+
+export function MarkerLabel(props: MarkerLabelProps): React.ReactNode {
+    const componentSize = useComponentSize();
```
rubenthoms: I am wondering why this is the only sub-component that uses the component size context directly instead of having a prop for size. All other sub-components get the size value drilled down. Is this by design?

rubenthoms (approved):
Looks good. Two things to consider: 1) clearing the input makes the slider jump to min/max (as discussed) and 2) why is the MarkerLabel component using the component size context directly.

frontend/src/lib/components/Slider/_components/markerLabel.tsx
Anders2303: No reason, think I was supposed to make it a prop, but just forgot; changed it to a prop now

PR #1821, opened 2026-07-09
[Lib] Menu item tones + submenu chevron alignment

Anders2303:
image
image

* Added a "tone" prop to all shared menu items
  * Supports `"neutral" | "warning" | "danger"`
  * To keep the shared logic with MenuItemContent, I reworked the component to use Base-UI's [`useRender`](https://base-ui.com/react/utils/use-render) util
* Fixed the sub-menu's chevron not being right-aligned


frontend/src/styles/components/menu.css
```diff
@@ -15,6 +15,22 @@
     }
 
     .menu__item {
+        &[data-tone="warning"] {
+            .menu__toggle_indicator {
+                @apply data-disabled:text-fill-disabled;
```
rubenthoms: You could consider using the same `not-data-disabled` pattern here as below in order to avoid having to repeat `data-disabled:text-fill-disabled` for all tones.

rubenthoms (approved):
One change to consider, otherwise LGTM - approved

rubenthoms (approved):

PR #1828, opened 2026-07-28
Fixed SeismicFenceMeshLayer crashing if properties contains NaN

Anders2303:
When a wellbore fence gets extended far enough values in the properties array might become NaN. This was observed on drogon_horizontal, with an extension length of 1500. _I assume it just happens because the fence goes outside of the simulated area?_


jorgenherje (approved):
See copilot comment

jorgenherje (approved):

PR #1837, opened 2026-07-31
[2D/3D viewer]: Improvements to polyline editing UX

### LINKED ISSUE #1619: Require polyline to consist of at least two points before being able to save polyline
...in order to not introduce error messages like "At least 2 control points are required" in modules consuming polylines (like Intersection).

The error message is quite clear in frontend in modules consuming polylines, but it is likely better to avoid module/crash error, and fix it at the source (i.e. never let the user incorrectly save polylines with only 1 point - both when new polylines are created and when editing existing ones).

rubenthoms:
Closes #1619.


frontend/src/modules/_shared/utils/subsurfaceViewer/PolylinesPlugin.tsx
Anders2303: One thought I had; we currently handle keyboard shortcuts for quitting and deleting; wouldn't it also make sense to add one for saving? Either as "ctrl+s" or "enter"?

Anders2303 (approved):
LGTM, good to merge. Noted one optional thing

frontend/src/modules/_shared/utils/subsurfaceViewer/PolylinesPlugin.tsx
rubenthoms: Added save-on-enter.

PR #1838, opened 2026-07-31
Subsurface-viewer: Fixed incorrect controller prop when toggling panning

Anders2303:
Resolves #1836 

When toggling pan via the `DeckGlInstanceManager`, we applied the changes direction to the `deck.controller` prop, this prop is just intended as a shorthand when using a default deck.gl view; so it only applies to the *first* view. Subsurface also doesn't account for this prop, meaning the double-click zoom was forced to `true`

This PR correctly applies the panning-option to each view via `DeckGlInstanceManager.makeDeckGlComponentProps()`

### DIFF
--- a/frontend/src/modules/_shared/utils/subsurfaceViewer/DeckGlInstanceManager.ts
+++ b/frontend/src/modules/_shared/utils/subsurfaceViewer/DeckGlInstanceManager.ts
@@ -99,6 +99,8 @@ export class DeckGlInstanceManager implements PublishSubscribe<DeckGlInstanceMan
     private _contextMenu: ContextMenu | null = null;
     private _verticalScale: number = 1;
 
+    private _panEnabled: boolean = true;
+
     private _hiddenViews: View[] = []; // plugin registered
     private _hiddenViewStatePatch: Record<string, any> = {};
     private _layerFilterWrappers: ((prev?: any) => any)[] = [];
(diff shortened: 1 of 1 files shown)


frontend/src/modules/_shared/utils/subsurfaceViewer/DeckGlInstanceManager.ts
```diff
@@ -383,6 +369,10 @@ export class DeckGlInstanceManager implements PublishSubscribe<DeckGlInstanceMan
                 viewports: (props.views?.viewports ?? []).map((viewport) => ({
                     ...viewport,
                     layerIds: [...(viewport.layerIds ?? []), ...pluginLayerIds],
+                    controller: {
```
rubenthoms, line 376: This is unconditionally overwriting the previous controller state - consider using a spread operator in order to keep the previous state.

rubenthoms (changes requested):
All in all, this looks good and makes use of the relatively new controller property which is the correct way to adjust this class. There are three changes that should be made:
1) Consider renaming `_panEnabled` to `_dragInteractionEnabled` (function names should be changed as well)
2) `disabledPanning`/`enablePanning` should have a guard in order to avoid unnecessary redraws
3) The controller state in the viewports should not be overridden unconditionally - currently, we have no component setting the controller property, but this makes these changes more future-proof.

rubenthoms (approved):

PR #1841, opened 2026-08-04
Subsurface Viewer: Rework info-pick to hoverdata transformation logic + small fixes

Anders2303:
## Changes
* Changes the hover data transformation function to be mapped and defined further up, and makes it easier to define custom transformations for modules that have their own custom layers
* Tweaks the multi-view picking to immediately transform picked coordinates to normal space
  * Other subsurface callbacks transform before emitting, so this the multi-pick more consistent
* Made hover effects appear on `onPickingInfoChange` so we catch hover data from all views
  * _This also means we emit hover data when the view is clicked, do we want this?_

## Example
As an example, adding hover output for the `SeismicFenceMeshLayer` would look like this:

```ts
//...modules/3DViewer/view/components/VisualizationAssemblerWrapper.tsx
import { makeHoverTransformationLookup } from "@modules/_shared/utils/subsurfaceViewer/hoverTransformations";

// ...

function getHoverDataFromSeismicFencePick(info: SeismicFenceMeshLayerPickingInfo): Partial<HoverData> {
    let hoverdata: HoverData= {}
    // ... Do any computations needde
    return hoverData
}

const CUSTOM_HOVER_DATA_TRANSFORMATIONS = makeHoverTransformationLookup([
    SeismicFenceMeshLayer,
    getHoverDataFromSeismicFencePick,
]);

export function VisualizationAssemblerWrapper(props: VisualizationAssemblerWrapperProps): React.ReactNode {
    //...
    return (
        <DpfSubsurfaceViewerWrapper
            {...}
            customHoverDataTransformations={CUSTOM_HOVER_DATA_TRANSFORMATIONS}
        />
    );
}

```


frontend/src/modules/_shared/utils/subsurfaceViewer/hoverTransformations.ts
```diff
@@ -0,0 +1,69 @@
...
+import {
+    getWellFeatureFromSubLayerData,
+    sanitizeMdReadout,
+    type ExtendedWellFeature,
+    type LayerPickInfoWithReadout,
+} from "@modules/_shared/utils/subsurfaceViewerLayers";
+
+export type HoverDataTransformation = <TInfo extends PickingInfo>(layerInfo: TInfo) => Partial<HoverData>;
+export type LayerCtor = abstract new () => Layer<any>;
+export type HoverTransformTuple = [LayerCtor, HoverDataTransformation];
```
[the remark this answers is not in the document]
Anders2303: The class ctor is only used as a key for looking up the transformer functions. Is there a better typing to use?

frontend/src/modules/_shared/utils/subsurfaceViewerLayers.ts
```diff
@@ -336,3 +254,15 @@ export function getGeoWellFeaturePath(wellFeature: BaseWellFeature): GeoJSON.Pos
...
     return wellFeature.geometry.geometries.find((g) => g.type === "LineString")?.coordinates ?? [];
 }
+
+export function getScaledCoordinate(coordinate: number[], verticalScale: number): number[] {
+    if (coordinate.length !== 3) return coordinate;
+
+    return [coordinate[0], coordinate[1], coordinate[2] * verticalScale];
+}
+
+export function getUnScaledCoordinate(coordinate: number[], verticalScale: number): number[] {
```
rubenthoms: Nitpick: `getUnscaled...`

frontend/src/modules/_shared/utils/subsurfaceViewer/hoverTransformations.ts
```diff
@@ -0,0 +1,69 @@
...
+export type LayerTransformationLookupMap = ReadonlyMap<LayerCtor, HoverDataTransformation>;
+
+/** Utility to create a read-only look-up for one or more picking transformations */
+export function makeHoverTransformationLookup(
+    ...hoverTransformations: HoverTransformTuple[]
+): LayerTransformationLookupMap {
+    return new Map([...hoverTransformations]);
+}
+
+export function getWellboreData(wellInfo: LayerPickInfoWithReadout<ExtendedWellFeature>): Partial<HoverData> {
```
rubenthoms: Consider reflecting the file's name in the functions - e.g. `transformWellboreToHoverData`. See others.

frontend/src/modules/_shared/components/SubsurfaceViewer/_components/HoverVisualizationWrapper.tsx
```diff
@@ -121,40 +118,43 @@ export function HoverVisualizationWrapper(props: HoverVisualizationWrapperProps)
...
     const handleViewportHover = React.useCallback(function handleViewportHover(viewport: ViewportType | null) {
         setCurrentlyHoveredViewport(viewport?.id ?? null);
     }, []);
 
     const handlePickingInfoChange = React.useCallback(
         function handlePickingInfoChange(newPickingInfoPerView: Record<string, PickingInfo[]>) {
-            // Convert to unscaled coordinates at the time of picking
-            // This ensures coordinates stay correct when verticalScale changes later
-            const unscaled: Record<string, [number, number, number][]> = {};
+            const coordPerView: Record<string, [number, number, number][]> = {};
```
rubenthoms: Nitpick: coordinates is always plural as soon as you talk about more than just one number.

frontend/src/modules/_shared/components/SubsurfaceViewer/_components/HoverVisualizationWrapper.tsx
```diff
@@ -121,40 +118,43 @@ export function HoverVisualizationWrapper(props: HoverVisualizationWrapperProps)
...
+                );
+
+                if (!transformationFunc) {
+                    return acc;
+                }
+
+                const hoverData = transformationFunc(info);
+
+                return { ...acc, ...hoverData };
+            }, {});
```
rubenthoms, line 165: Hm, I think we have to be a bit careful here with keeping the original pick-order (nearest first). `allPickingInfo` preserves that order, but on line 152 the previous values might get overridden. Reconsider this approach, reversing order might be sufficient.

rubenthoms (changes requested):
All in all, this looks very good :+1: One major issue to reconsider, otherwise mostly minor ones.

frontend/src/modules/_shared/utils/subsurfaceViewer/hoverTransformations.ts
Anders2303: Is changing them like this ok?
`getWellboreData` -> `transformToWellboreHoverData`
`getWorldPosition` -> `transformToWorldPosHoverData`

 I want to make it clear that the function makes partial hover data, and not all applicable data (for example, wellbore pick does have coordinates. but we omit `HoverTopic.WORLD_POS_UTM`)

rubenthoms: Absolutely

frontend/src/modules/_shared/components/SubsurfaceViewer/_components/HoverVisualizationWrapper.tsx
Anders2303: Reversing should be fine. I also [added some logic](https://github.com/equinor/webviz/pull/1841/commits/47cc88065a941800722b79675a22c17a9e3c3500) to make it so we also prioritize data from the hovered/clicked viewport

frontend/src/modules/_shared/utils/subsurfaceViewerLayers.ts
```diff
@@ -261,7 +261,7 @@ export function getScaledCoordinate(coordinate: number[], verticalScale: number)
     return [coordinate[0], coordinate[1], coordinate[2] * verticalScale];
 }
 
-export function getUnScaledCoordinate(coordinate: number[], verticalScale: number): number[] {
+export function getUnscaledCoordinate(coordinate: number[], verticalScale: number): number[] {
```
rubenthoms: Nitpick: `getUnscaledCoordinates` and ` coordinates`

rubenthoms (approved):
LGTM :+1: Two leftover uses of `coordinate` instead of ` coordinates` you might want to change before merging in

PR #1844, opened 2026-08-06
Feat: Startpage changelog

Anders2303:
Adds a mechanism to show a changelog on the start-page. When a new, unread changelog is detected, a modal will automatically open.

* Adds `react-markdown` as a project dependency
  * Only basic markdown language is supported per now (no video, tables, and so forth)
* Adds a `/docs` folder at the project root; we should use this folder going forward to store documentation, like the changelog
  * The folder is copied from the root to the frontend container when starting the docker server, so VS Code might incorrectly complain about document imports being invalid...
  * Additionally, Vite is unable to notice file changes in `/docs` so markdown changes are only visible after rebuilding
* When opening the start page, the file `/docs/WEBVIZ_CHANGELOG.MD` is loaded, and the release (as defined by the `% release: x` line at the top of the file) is compared to the `lastSeenChangelog` value in the users local-storage

## Usage
To test the feature locally, modify the `WEBVIZ_CHANGELOG.MD` file. Increment the version number to make it open automatically

## Note
Per now, the "see full changelog" link in the modal is hardcoded to the changelog file location in the github project; specifically pointing to `main` branch. We should consider instead linking to an explicit release tag instead to ensure things are synced


docs/WEBVIZ_CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+% release: 0
```
anders-kiaer: Should we name this variable something else than `release`? E.g. `changelog_counter`?

We will release potentially to Radix on every commit, but not every commit will add a changelog entry (e.g. deps updates, and other changes not visible to the users).

docs/images/open_changelog_animated.avif
anders-kiaer, line 1: Remove before merge

docs/images/changelog_system.png
anders-kiaer, line 1: Remove before merge

frontend/vite.config.ts
```diff
@@ -67,6 +68,25 @@ export default defineConfig(() => {
...
+            if (id.endsWith(".md")) {
+                // Return the file content as a default export string
+                return {
+                    code: `export default ${JSON.stringify(code)}`,
+                    map: null,
+                };
+            }
+        },
+    };
+}
```
anders-kiaer: Is it enough with Vite's native `raw` support here? https://vite.dev/guide/assets#importing-asset-as-string

```
import myMarkdown from './content.md?raw';

console.log(myMarkdown);
```

Async variant:
```
const myMarkdown = (await import('./content.md?raw')).default;
```

anders-kiaer (approved):
A couple of suggestions/questions - otherwise LGTM 👍🙂

frontend/vite.config.ts
Anders2303: Yeah, the plugin was intended to replace the native raw, just so you don't need to worry remembering that every time. I can change it to just use the native one if thats preferable

PR #1831, opened 2026-07-29
[HoverService] Synced hover between Polylines, seismic fence layer and intersection modules

Anders2303:
Adds a "position along fence" hover topic to the hover service, replacing the existing "position along polyline" topic.

This topic is now used to show synchronized position hover on:
* The Intersection module
* The 3D-viewer's SeismicFence layer
* Polylines overview

Polylines were the odd one out, so I made the hover payload only worry about the XY plane, and the polyline layer is responsible for converting the fence's position-along to it's own explicit length.

PR #1850, opened 2026-08-13
[Intersection]: Added multi-view readout to intersection via fence hover topic

rubenthoms:
Added: Multi-view and multi-module readout for same fence in Intersection module.
Fixed: X,Y coordinates readout was not extrapolated after leaving wellbore fence range.

---

Waiting for #1831 as it already introduces some parts of this PR.


anders-kiaer:
Blocked by #1831.

PR #1858, opened 2026-08-19
Inplace volumes waterfall

### LINKED ISSUE #1808: Volume waterfall (ensemble comparison)

HansKallekleiv:
Closes #1808


frontend/src/modules/InplaceVolumesComparison/preview.tsx
rubenthoms: Could be better aligned with the other preview images.

frontend/src/modules/InplaceVolumesComparison/settings/atoms/derivedAtoms.ts
```diff
@@ -0,0 +1,300 @@
...
+    const sourceTableDefinitions = get(sourceTableDefinitionsAtom);
+    const sourceSpecs = [
+        { ensembleIdent: get(selectedReferenceEnsembleIdentAtom).value, definition: sourceTableDefinitions.reference },
+        {
+            ensembleIdent: get(selectedComparisonEnsembleIdentAtom).value,
+            definition: sourceTableDefinitions.comparison,
+        },
+    ];
+
+    const definitionsPerSource: TableDefinitionsPerEnsemble[] = [];
```
rubenthoms: The variable names used here are a bit misleading when reading the code. It states `definitionSPerSource` although there can only be ONE definition per source.

frontend/src/modules/InplaceVolumesComparison/settings/atoms/derivedAtoms.ts
```diff
@@ -0,0 +1,300 @@
...
+    for (const spec of sourceSpecs) {
+        if (!spec.ensembleIdent || !spec.definition) {
+            continue;
+        }
+        definitionsPerSource.push({ ensembleIdent: spec.ensembleIdent, tableDefinitions: [spec.definition] });
+    }
+
+    const tableNamesFilter = Array.from(
```
rubenthoms: `tableNamesFilter` is also misleading as it suggests that it is holding a function, not an `allowedTableNames` list.

frontend/src/modules/InplaceVolumesComparison/settings/atoms/derivedAtoms.ts
```diff
@@ -0,0 +1,300 @@
...
+        if (referenceOnlyValues.length > 0 || comparisonOnlyValues.length > 0) {
+            differences.push({ indexColumn, missingFrom: null, referenceOnlyValues, comparisonOnlyValues });
+        }
+    }
+
+    return differences;
+});
+
+/** True when the user has opted into filtering on the intersection of differing index values. */
+export const isIndexValueIntersectionEnabledAtom = atom(
```
rubenthoms: This does not need to be an atom - could be a simple helper function with `get` as parameter. No need to cache its result value.

frontend/src/modules/InplaceVolumesComparison/settings/atoms/persistableFixableAtoms.ts
```diff
@@ -0,0 +1,118 @@
...
+import {
+    availableComparisonTableNamesAtom,
+    availableIndicesWithValuesAtom,
+    availableReferenceTableNamesAtom,
+    availableResultNamesAtom,
+    commonIndicesWithValuesAtom,
+} from "./derivedAtoms";
+import { tableDefinitionsQueryAtom } from "./queryAtoms";
+
+function computeTableDefinitionsQueryDependenciesState({ get }: { get: Getter }): "error" | "loading" | "loaded" {
```
rubenthoms: In order to keep in sync it makes sense to export the `PersistableAtomDependenciesState` from `atomUtils.ts` and use it here instead of defining your own return type.

frontend/src/modules/InplaceVolumesComparison/settings/atoms/queryAtoms.ts
```diff
@@ -0,0 +1,52 @@
...
+export const tableDefinitionsQueryAtom = atomWithQueries((get) => {
+    const referenceEnsembleIdent = get(selectedReferenceEnsembleIdentAtom).value;
+    const comparisonEnsembleIdent = get(selectedComparisonEnsembleIdentAtom).value;
+
+    const ensembleIdents: RegularEnsembleIdent[] = [];
+    for (const ensembleIdent of [referenceEnsembleIdent, comparisonEnsembleIdent]) {
+        if (ensembleIdent && !ensembleIdents.some((existing) => existing.equals(ensembleIdent))) {
+            ensembleIdents.push(ensembleIdent);
+        }
+    }
```
rubenthoms: A bit of an overkill for only two ensembles? Will the number of ensembles ever change? Otherwise, just add the first one unconditionally and use a simple if-equals check for the second.

frontend/src/modules/InplaceVolumesComparison/view/atoms/queryAtoms.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export const isWaterfallComputableAtom = atom((get) => {
+    return (
+        get(areSourcesDistinctAtom) &&
+        get(areSelectedTablesComparableAtom) &&
+        get(areSelectedIndicesWithValuesValidAtom) &&
+        get(waterfallFactorSpecAtom) !== null &&
+        get(waterfallSourcesAtom) !== null
+    );
+});
```
rubenthoms: These are no query atoms.

frontend/src/modules/InplaceVolumesComparison/view/atoms/queryAtoms.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    if (!referenceEnsembleIdent || !comparisonEnsembleIdent || !referenceTableName || !comparisonTableName) {
+        return null;
+    }
+
+    return {
+        reference: { ensembleIdent: referenceEnsembleIdent, tableName: referenceTableName },
+        comparison: { ensembleIdent: comparisonEnsembleIdent, tableName: comparisonTableName },
+    };
+});
```
rubenthoms: Why not reusing `sourceTableDefinitionsAtom` from the settings? Then you could also remove all single-value atoms in the interface.

frontend/src/modules/InplaceVolumesComparison/settings/atoms/persistableFixableAtoms.ts
```diff
@@ -0,0 +1,118 @@
...
+        // SELECT_ALL fixup runs and the column is not left as an empty (everything-filtered) filter.
+        const coversAllAvailableColumns = availableIndicesWithValues.every((available) =>
+            value.some((selected) => selected.indexColumn === available.indexColumn),
+        );
+        return (
+            value.length > 0 &&
+            coversAllAvailableColumns &&
+            isSelectedIndicesWithValuesValidSubset(value, availableIndicesWithValues)
+        );
+    },
```
rubenthoms: If there is no intersection between the two sources this would become `[]` and `isValidFunction` would return `false`, calling the `fixupFunction` which again returns `[]`. It seems `[]` is a valid selection here but it gets flagged when persisted, resulting in a warning message which should arguably not be there if the selection was not flagged before.

frontend/src/modules/InplaceVolumesComparison/settings/settings.tsx
```diff
@@ -0,0 +1,416 @@
...
+    const isSameSourceSelectedTwice = Boolean(
+        referenceEnsembleIdent.value &&
+        comparisonEnsembleIdent.value &&
+        selectedReferenceTableName.value &&
+        selectedComparisonTableName.value &&
+        !areSourcesDistinct,
+    );
+
+    const referenceEnsembleAnnotations: SettingAnnotation[] = isSameSourceSelectedTwice
+        ? [...persistedReferenceEnsembleAnnotations, { type: "error", message: "Must differ from the comparison" }]
```
rubenthoms: Consider moving the annotation to the comparison ensemble.

frontend/src/modules/InplaceVolumesComparison/settings/settings.tsx
```diff
@@ -0,0 +1,416 @@
...
+    const comparisonSourceAnnotations: SettingAnnotation[] = [
+        ...comparisonEnsembleAnnotations,
+        ...comparisonTableNameAnnotations,
+    ];
+
+    if (areSourcesDistinct && !areSelectedTablesComparable) {
+        statusWriter.addWarning("The selected table sources share no index columns and are not comparable.");
+    }
+
+    const isIndexValueIntersectionEnabled = selectedIndexValueCriteria === IndexValueCriteria.ALLOW_INTERSECTION;
```
rubenthoms: You already have an atom for this which I suggested to become a helper function. However, if you reuse it here, you could leave it being an atom.

frontend/src/modules/InplaceVolumesComparison/settings/settings.tsx
```diff
@@ -0,0 +1,416 @@
...
+                    >
+                        <>
+                            <EnsembleDropdown
+                                ensembles={ensembleSet.getRegularEnsembleArray()}
+                                value={referenceEnsembleIdent.value}
+                                ensembleRealizationFilterFunction={ensembleRealizationFilterFunction}
+                                onValueChange={setReferenceEnsembleIdent}
+                            />{" "}
+                            <Combobox
+                                startAdornment={<span>Table source:</span>}
```
rubenthoms: You must not use the `startAdornment` for a label. It is not meant to be used this way and is also unconditionally replaced by an error icon. Consider changing the label of the field to `Reference source` (and `Comparison source` respectively) and giving each input their own `aria-label`.

frontend/src/modules/InplaceVolumesComparison/settings/settings.tsx
```diff
@@ -0,0 +1,416 @@
...
+                    >
+                        <>
+                            <EnsembleDropdown
+                                ensembles={ensembleSet.getRegularEnsembleArray()}
+                                value={comparisonEnsembleIdent.value}
+                                ensembleRealizationFilterFunction={ensembleRealizationFilterFunction}
+                                onValueChange={setComparisonEnsembleIdent}
+                            />{" "}
+                            <Combobox
+                                startAdornment={<span>Table source:</span>}
```
rubenthoms: See above.

frontend/src/modules/InplaceVolumesComparison/view/hooks/useBuildWaterfallPlot.ts
```diff
@@ -0,0 +1,383 @@
...
+} from "../utils/computeVolumeChangeDecomposition";
+import { findTableDataForSource, makeSourceLabels } from "../utils/waterfallSources";
+
+export type WaterfallMessage = {
+    text: string;
+    /** "info" for an incomplete selection the user still has to finish, "error" for a real failure. */
+    severity: "info" | "error";
+};
+
+export interface WaterfallResult {
```
rubenthoms: This name can be very confusing and misleading. Consider `UseBuildWaterfallPlotResult`

frontend/src/modules/InplaceVolumesComparison/view/hooks/useBuildWaterfallPlot.ts
```diff
@@ -0,0 +1,383 @@
...
+} from "../atoms/queryAtoms";
+import { buildWaterfallPlot, type WaterfallGroupDecomposition } from "../utils/buildWaterfallPlot";
+import {
+    computeVolumeChangeDecomposition,
+    getRequiredFluidForWaterfallTarget,
+    isWaterfallTargetResultName,
+} from "../utils/computeVolumeChangeDecomposition";
+import { findTableDataForSource, makeSourceLabels } from "../utils/waterfallSources";
+
+export type WaterfallMessage = {
```
rubenthoms: `WaterfallMessage` sounds very weird - consider finding a better name or skip it entirely and bake it into the result type - no consumers anyway.

frontend/src/modules/InplaceVolumesComparison/view/hooks/useBuildWaterfallPlot.ts
```diff
@@ -0,0 +1,383 @@
...
+    text: string;
+    /** "info" for an incomplete selection the user still has to finish, "error" for a real failure. */
+    severity: "info" | "error";
```
rubenthoms: `severity` seems to be the wrong term for your case. Consider calling it `reason: "incomplete-selection" | "failure"` or something similar.

frontend/src/modules/InplaceVolumesComparison/view/hooks/useBuildWaterfallPlot.ts
```diff
@@ -0,0 +1,383 @@
...
+};
+
+export interface WaterfallResult {
+    plots: React.ReactNode | null;
+    isFetching: boolean;
+    /** User-facing message when the waterfall cannot be shown. */
+    message: WaterfallMessage | null;
+    /** Non-blocking notes shown alongside a rendered plot. */
+    warnings: string[];
```
rubenthoms: A warning is a message. So this is a bit confusing. Consider being more explicit about what things really are: e.g. `placeholderMessage` vs. `nonBlockingWarnings`. Even better: consider returning the no-data/incomplete-selection placeholder (currently in `view` with `<ContentInfo>` instead of the plots (you are already returning `React.ReactNode`) and renaming the field to `errorMessage: string | null`.

frontend/src/modules/InplaceVolumesComparison/view/hooks/useBuildWaterfallPlot.ts
```diff
@@ -0,0 +1,383 @@
...
+    }
+    if (!isWaterfallTargetResultName(resultName)) {
+        return makeErrorResult("Neither STOIIP nor GIIP is available for the selected tables.");
+    }
+    if (!spec) {
+        return makeErrorResult(
+            "The volume columns the decomposition is built from (BULK, PORV, HCPV) are not available for the selected table.",
+        );
+    }
+    if (!isComputable) {
```
rubenthoms: Currently unreachable as all conditions are already checked before in this code. However, this would fail silently without any user-facing message.

frontend/src/modules/InplaceVolumesComparison/view/hooks/useBuildWaterfallPlot.ts
```diff
@@ -0,0 +1,383 @@
...
+            decomposition,
+            uncertainty:
+                referenceStatistics.targetBand && comparisonStatistics.targetBand
+                    ? {
+                          reference: referenceStatistics.targetBand,
+                          comparison: comparisonStatistics.targetBand,
+                      }
+                    : null,
+        });
+    }
```
rubenthoms: There is a lot of rather expensive code here that is also called whenever some irrelevant props are changing (e.g. width/height). Consider memoizing the results of these functions.

frontend/src/modules/InplaceVolumesComparison/view/hooks/useBuildWaterfallPlot.ts
```diff
@@ -0,0 +1,383 @@
...
+    const referenceStatisticsByGroup = extractRequiredStatisticsByGroup(
+        referenceTableData,
+        spec.requiredResultNames,
+        spec.target,
+        requiredFluid,
+        subplotByIndex,
+    );
+
+    if (!comparisonStatisticsByGroup || !referenceStatisticsByGroup) {
+        return makeErrorResult(noDataMessage);
```
rubenthoms: Isn't this very misleading for the user? For instance, the requested group-by (subplot-by) column might just be missing from the statistical data.

frontend/src/modules/InplaceVolumesComparison/view/hooks/useBuildWaterfallPlot.ts
```diff
@@ -0,0 +1,383 @@
...
+                    ? {
+                          reference: referenceStatistics.targetBand,
+                          comparison: comparisonStatistics.targetBand,
+                      }
+                    : null,
+        });
+    }
+
+    if (groupDecompositions.length === 0) {
+        return makeErrorResult("The waterfall could not be computed for the selected data.");
```
rubenthoms: In some cases you could already be way more specific about the actual error and make it actionable for the user. For instance, if the two sides (ref/comp) share zero common groups which you already compute on lines 296-301.

frontend/src/modules/InplaceVolumesComparison/settings/settings.tsx
```diff
@@ -0,0 +1,416 @@
...
+                        </>
+                    </Setting.Field>
+
+                    {isCrossTableComparison && (
+                        <Banner tone="warning" layoutClassName="col-span-3">
+                            <strong>Comparing different table sources.</strong> Differences in BULK and PORO between
+                            table sources reflect gridding and upscaling as well as any reservoir change, so the factor
+                            contributions are not a like-for-like comparison.
+                            {isSingleEnsembleComparison && " Both sides use the same ensemble."}
+                        </Banner>
```
rubenthoms: Consider x-margin for the banners to align them vertically with the rest of the setting components.

frontend/src/modules/InplaceVolumesComparison/view/utils/buildWaterfallPlot.tsx
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * Hover text per bar. The endpoint bars also carry the ensemble's uncertainty band as text: the
+ * within-ensemble spread is typically far wider than the change being decomposed, so drawing it on
+ * the same axis would dwarf the factor bars.
+ */
+function makeBarHoverTexts(group: WaterfallGroupDecomposition, displayLabels: string[]): string[] {
+    const { bars } = group.decomposition;
+    const barTexts = makeBarTexts(group.decomposition);
+
+    return bars.map((bar, index) => {
```
rubenthoms: Mark `bar` as unused: `_`.

frontend/src/modules/InplaceVolumesComparison/view/utils/buildWaterfallPlot.tsx
```diff
@@ -0,0 +1,174 @@
...
+                range: makeYAxisRange(group.decomposition),
+                gridcolor: "lightgrey",
+            },
+            [`xaxis${axisIndex}`]: {
+                type: "category",
+            },
+        });
+    });
+
+    figure.updateLayout({ showlegend: false, plot_bgcolor: "white" });
```
rubenthoms: Unnecessary to set the bgcolor to `white`? We haven't done that anywhere else.

frontend/src/modules/InplaceVolumesComparison/view/utils/buildWaterfallPlot.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * Render one volume-change waterfall per group (e.g. per REGION) in a subplot grid, using Plotly's
+ * native "waterfall" trace. A single group yields a single, ungrouped waterfall.
+ */
+export function buildWaterfallPlot(
+    groups: WaterfallGroupDecomposition[],
+    options: WaterfallPlotOptions,
+): React.ReactNode {
+    const numSubplots = Math.max(groups.length, 1);
```
rubenthoms: Should we return `null` when `groups.length === 0`?

frontend/src/modules/InplaceVolumesComparison/view/utils/buildWaterfallPlot.tsx
```diff
@@ -0,0 +1,174 @@
...
+            },
+            [`xaxis${axisIndex}`]: {
+                type: "category",
+            },
+        });
+    });
+
+    figure.updateLayout({ showlegend: false, plot_bgcolor: "white" });
+
+    return <Plot data={figure.makeData() as Data[]} layout={figure.makeLayout()} />;
```
rubenthoms: Type-cast really necessary?

frontend/src/modules/InplaceVolumesComparison/view/utils/computeVolumeChangeDecomposition.ts
```diff
@@ -0,0 +1,253 @@
...
+function computeFactorValue(factor: WaterfallFactor, means: Map<string, number>): number | null {
+    const numerator = means.get(factor.numeratorResultName);
+    if (numerator === undefined) {
+        return null;
+    }
+    if (factor.denominatorResultName === null) {
+        return numerator;
+    }
+    const denominator = means.get(factor.denominatorResultName);
+    if (denominator === undefined || denominator === 0) {
```
rubenthoms: Maybe we should also consider guarding against near-zero values as they might cause a wildly dominant bar or absurd percentage in the waterfall. We should maybe handle that as numerically unstable?

frontend/src/modules/InplaceVolumesComparison/view/utils/computeVolumeChangeDecomposition.ts
```diff
@@ -0,0 +1,253 @@
...
+    if (factor.dividesVolume) {
+        if (comparisonValue === 0) {
+            return null;
+        }
+        return referenceValue / comparisonValue - 1;
+    }
+
+    if (referenceValue === 0) {
+        return null;
+    }
```
rubenthoms: See above.

frontend/src/modules/InplaceVolumesComparison/view/utils/computeVolumeChangeDecomposition.ts
```diff
@@ -0,0 +1,253 @@
...
+): WaterfallFactorSpec | null {
+    if (!isWaterfallTargetResultName(targetResultName)) {
+        return null;
+    }
+
+    const available = new Set(availableResultNames);
+    const saturationLabel = SATURATION_LABEL_BY_TARGET[targetResultName];
+    const fvfLabel = FVF_LABEL_BY_TARGET[targetResultName];
+
+    // NTG and PORO_NET are advertised exactly when NET is present, and NET is what the split needs.
```
rubenthoms: If this was true, why checking for "NTG" / "PORO_NET" then?

frontend/src/modules/InplaceVolumesComparison/view/utils/buildWaterfallPlot.tsx
```diff
@@ -0,0 +1,174 @@
...
+    return {
+        type: "waterfall",
+        orientation: "v",
+        measure: bars.map((bar) => bar.measure),
+        x: labels,
+        y: bars.map((bar) => bar.value),
+        text: makeBarTexts(group.decomposition),
+        textposition: "outside",
+        textfont: { size: 11 },
+        hovertext: makeBarHoverTexts(group, labels),
```
rubenthoms: Minor: `makeBarHoverTexts` could reuse the result of `makeBarTexts`.

frontend/src/modules/InplaceVolumesComparison/view/utils/waterfallBarPresentation.ts
```diff
@@ -0,0 +1,33 @@
...
+ * endpoint bars, which are volumes rather than changes, and for a zero previous cumulative, where
+ * the percentage is undefined rather than meaningfully zero.
+ */
+export function computeBarChangePercent(bars: WaterfallBar[], index: number): number | null {
+    const bar = bars[index];
+    if (bar.measure === "absolute") {
+        return null;
+    }
+    const previousCumulative = bars[index - 1]?.cumulative ?? 0;
+    return previousCumulative !== 0 ? (100 * bar.value) / previousCumulative : null;
```
rubenthoms: Same near-zero consideration as above.

frontend/src/modules/InplaceVolumesComparison/view/utils/waterfallPlotLayout.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @returns The number of rows and columns to use.
+ */
+export function calcNumRowsAndCols(numSubplots: number): { numRows: number; numCols: number } {
+    if (numSubplots < 1) {
+        return { numRows: 1, numCols: 1 };
+    }
+    const numRows = Math.ceil(Math.sqrt(numSubplots));
+    const numCols = Math.ceil(numSubplots / numRows);
+    return { numRows, numCols };
+}
```
rubenthoms: This is a function we already have implemented a couple of times (see PlotBuilders). Might be worth it to make a shared util before implementing another copy.

frontend/src/modules/InplaceVolumesComparison/view/utils/waterfallSources.ts
```diff
@@ -0,0 +1,43 @@
+import type { RegularEnsembleIdent } from "@framework/RegularEnsembleIdent";
+
+/** One side of the comparison: a table within an ensemble. */
+export type WaterfallSource = {
+    ensembleIdent: RegularEnsembleIdent;
+    tableName: string;
+};
+
+type TableDataForSource = {
+    ensembleIdent: { equals: (other: RegularEnsembleIdent) => boolean };
```
rubenthoms: The only call-site of this is using `RegularEnsembleIdent` so I don't quite see how this abstraction is defendable.

frontend/src/modules/InplaceVolumesComparison/view/utils/waterfallSources.ts
```diff
@@ -0,0 +1,43 @@
+import type { RegularEnsembleIdent } from "@framework/RegularEnsembleIdent";
+
+/** One side of the comparison: a table within an ensemble. */
+export type WaterfallSource = {
+    ensembleIdent: RegularEnsembleIdent;
+    tableName: string;
+};
+
+type TableDataForSource = {
```
rubenthoms: This name is very misleading - I don't see any "data".

frontend/src/modules/_shared/InplaceVolumes/TableDefinitionsAccessor.ts
```diff
@@ -134,11 +146,9 @@ export class TableDefinitionsAccessor {
                         indexColumn: index,
                         values: valuesIntersection,
                     });
-                } else if (!isEqual([...indexWithValues.values].sort(), [...currentIndexWithValues.values].sort())) {
-                    // Compare sorted copies so equality is order-insensitive without changing the
-                    // backend-provided order used by selectors and plots.
+                } else if (!areValuesEqual) {
+                    // Tables are not comparable when index values are not equal
                     this._tablesNotComparable = true;
-                    break;
```
rubenthoms: Have you checked the implications of removing the `break` for the other three consumers?

rubenthoms (changes requested):
All in all, this looks good :+1: A couple of changes to consider.

frontend/src/modules/InplaceVolumesComparison/view/hooks/useBuildWaterfallPlot.ts
```diff
@@ -211,173 +233,249 @@ export function useBuildWaterfallPlot(ensembleSet: EnsembleSet, width: number, h
...
-        return makeErrorResult(
-            "The volume columns the decomposition is built from (BULK, PORV, HCPV) are not available for the selected table.",
-        );
-    }
-    if (!isComputable) {
-        return makePendingResult(false);
-    }
+    // Everything below is independent of width/height, so it is memoized to avoid redoing the
+    // per-group statistics extraction and decomposition on every resize.
+    const computation = useMemo((): ComputationResult => {
```
rubenthoms: For such a large function you could improve traceability by giving it a name.

rubenthoms (approved):
Consider using a named function inside the `useMemo` hook to improve traceability. Otherwise, LGTM :+1:

PR #1861, opened 2026-08-19
Fixed outdated transitive deps in primary lockfile

Anders2303:
When updating depdencies in the library sub-projects, the new dependencies are not reflected in the primary project lockfile. This can be seen in #1738, where primary still maintained a pyarrow version of 23, as path dependencies get cached initially.

Going forward, we need to make sure to always regenerate lock files (`poetry lock`) in dependents when updating deps in the lib projects. I added a script to quickly do this for each pyproject, and added a lock check to the pipeline script


sigurdp (approved):
👍

PR #1871, opened 2026-08-25
Fixes to Inplace modules

HansKallekleiv:
- Use same number formatting in all visualizations
- Make it easier to visualize Bo/Bg by automatically adding FLUID as a grouping or disable with tooltip explanation
- Handle missing result columns from backend
- Preserve index value order from backend
- Replace result selection dropdown/select with combobox with descriptions


frontend/src/modules/InplaceVolumesNew/settings/settings.tsx
```diff
@@ -133,10 +142,16 @@ export function Settings(props: ModuleSettingsProps<Interfaces>): React.ReactNod
...
+                        {selectedFirstResultName.value} is calculated per fluid. Data is automatically grouped by
+                        FLUID. {!isRequiredFluidSelected && `Include ${requiredFluid} in the FLUID filter to see data.`}
+                    </Banner>
+                )}
                 {showFaciesFractionGroupingWarning && (
-                    <Banner tone="warning">
-                        <strong>Note:</strong> FACIES_FRACTION is only meaningful when FACIES is used as Subplot by or
-                        Color by; otherwise every fraction collapses to 1.
+                    <Banner layoutClassName="col-span-3" tone="warning">
+                        FACIES_FRACTION is only meaningful when FACIES is used as Subplot by or Color by; otherwise
```
rubenthoms: Use quotes? `FACIES_FRACTION is only meaningful when FACIES is used as "Subplot by" or "Color by"; ...`

frontend/src/modules/InplaceVolumesNew/view/utils/GroupedTableData.ts
rubenthoms: Should `buildColorMap` also match the display order? Currently, it still sorts alphabetically.

frontend/src/modules/_shared/InplaceVolumes/PlotBuilder.tsx
```diff
@@ -77,9 +82,9 @@ export class PlotBuilder {
 
                 figure.updateLayout({
                     // @ts-expect-error - Ignore string type of xAxisKey for oldLayout[xAxisKey]
-                    [xAxisKey]: { ...oldLayout[xAxisKey], ...this._axesOptions.x },
+                    [xAxisKey]: { ...oldLayout[xAxisKey], ...this._numberFormatAxisOptions, ...this._axesOptions.x },
                     // @ts-expect-error - Ignore string type of yAxisKey for oldLayout[yAxisKey]
-                    [yAxisKey]: { ...oldLayout[yAxisKey], ...this._axesOptions.y },
+                    [yAxisKey]: { ...oldLayout[yAxisKey], ...this._numberFormatAxisOptions, ...this._axesOptions.y },
```
rubenthoms: Are we sure we always want the same formatting for both axes or should we provide two distinct ones? Could lead to weird number formatting (like sub-1% values being displayed as 500m, for instance).

frontend/src/modules/_shared/InplaceVolumes/indexWithValuesUtils.ts
```diff
@@ -2,6 +2,11 @@ import type { InplaceVolumesIndexWithValues_api } from "@api";
 import type { FixupSelection } from "@lib/utils/fixupUserSelection";
 import { fixupUserSelection } from "@lib/utils/fixupUserSelection";
 
+export function orderSelectedIndexValues(selectedValues: string[], availableValues: string[]): string[] {
+    const selectedValueSet = new Set(selectedValues);
+    return availableValues.filter((value) => selectedValueSet.has(value));
+}
```
rubenthoms: Mismatch of function name with what it is actually doing? It is not sorting but filtering?

rubenthoms (changes requested):
LGTM :+1: Some small adjustments, then ready to merge

rubenthoms (approved):

PR #1874, opened 2026-08-27
[Framework]: Boundaries for GL context loss + adapters for all GL components

rubenthoms:
### Problem
Browsers cap the number of live WebGL/WebGPU contexts a page may hold. With many graphics-intensive views or tabs open, the browser silently drops a canvas' context, leaving the visualization frozen or blank with no indication of what happened or how to get it back.

### Solution
A new `GpuResourceBoundary` component that wraps a GL-backed renderer, detects context loss, shows an explanatory overlay with a **Restore** action, and retries automatically when the tab becomes visible again. Renderers plug in through a small `GpuResourceAdapter` contract, so the boundary stays library-agnostic.

---

### `GpuResourceBoundary` + the adapter contract
`frontend/src/framework/components/GpuResourceBoundary/`

- **`GpuResourceBoundary`** — overlay UX + recovery lifecycle. The wrapper element is `display: contents` while the context is healthy, so it adds no box to the layout / measurement tree; it only becomes a positioned, sized box while the overlay is shown (same element, no child remount).
- **`GpuResourceAdapter`** — the renderer bridge:
  - `connect({ onContextLost, onContextRestored })` → returns a detach function. Runs *after* the renderer mounts, so implementations must also probe for an already-lost context on connect and buffer a notification that arrives before it.
  - `restoreContext?()` → `boolean` — bring a *lost* context back in place (e.g. `WEBGL_lose_context.restoreContext()`). Returns `true` if a restore was started (expect `webglcontextrestored`), `false` if it couldn't.
  - `requestRender?()` — repaint the *current, valid* context. Never restores a lost one.
- **`GpuRecoveryStrategy`**:
  - `"redraw"` — on Restore the boundary calls `restoreContext()`, keeps the overlay until `webglcontextrestored`, then `requestRender()`. Falls back to `"remount"` if `restoreContext` is missing, reports it couldn't start, or `webglcontextrestored` doesn't arrive within a short timeout (`REDRAW_RESTORE_TIMEOUT_MS`) — a real eviction isn't always restorable in place, and then no event ever fires.
  - `"remount"` (default) — bumps a `key` so the children remount with a fresh canvas/context. An in-place `webglcontextrestored` is *not* trusted here (some renderers can't rebuild GPU resources on a restored context) — it also triggers a remount.
- Auto-recovery: retries when the tab becomes **visible** again after the context was lost while it was in the background. Keyed on visibility, not window focus, so two visible windows don't fight over a scarce context. The `connect` effect re-subscribes on `recoveryStrategy` changes.

### Reusable adapter building blocks
`GpuResourceBoundary/adapters/`

- **`createCanvasContextAdapter({ getCanvas, restoreContext?, requestRender?, isContextLost?, waitForCanvas? })`** — for any renderer drawing to a `<canvas>` with the standard `webglcontextlost`/`webglcontextrestored` events. Handles `preventDefault()`, async-canvas polling, and the already-lost probe. Forwards signals only — the boundary owns all repaint/restore decisions.
- **`createManualContextLossAdapter()`** — for renderers exposing only a single "context lost" callback and no restored signal (Plotly). Buffers a notification received before `connect`.

### Renderer-specific adapters
Kept next to their renderer so `@framework` stays free of deck.gl/Pixi deps.

- **`SubsurfaceViewer/_utils/deckGlGpuResourceAdapter.ts`** — `createDeckGlGpuResourceAdapter(deckGl)`. `waitForCanvas`, `isContextLost` (guarded on `deck.isInitialized`). No `restoreContext` → uses `"remount"`.
- **`EsvIntersection/esvIntersectionGpuResourceAdapter.ts`** — `createEsvIntersectionGpuResourceAdapter(controller)`. Provides `restoreContext` (via `WEBGL_lose_context`), `requestRender`, `isContextLost` → uses `"redraw"`, with remount as fallback when the extension is unavailable or the restore doesn't complete in time.

### Supporting changes

- **`useIsDocumentVisible()`** (new, `lib/hooks/`) — tracks tab visibility (`!document.hidden`) via `visibilitychange`. Deliberately *not* keyed on window focus.
- **`EsvIntersectionController`**:
  - New `isContextLost()` and `restoreContext(): boolean`; `requestRender()` is now just a repaint.
  - `destroy()` now explicitly tears down the Pixi `stage` + `renderer` (`PixiRenderApplication` has no `destroy()` of its own), releasing the WebGL context immediately instead of at GC — and does the same for the app created in the `initialize()`/`destroy()` race.
- **`DeckGlInstanceManager.setRef()`** — no-ops on an unchanged ref, triggers a redraw on a real change so subscribers rebuild their deck props.

### Consumers wired up

| Component | Renderer | Strategy |
|---|---|---|
| `SubsurfaceViewerWithCameraState` | deck.gl | `remount` |
| `EsvIntersection` | Pixi.js | `redraw` (→ remount fallback) |
| `Plot` | Plotly | `remount` |

- **`SubsurfaceViewerWithCameraState`** exposes `onDeckGlInstanceChange`; `ReadoutWrapper` feeds it into `DeckGlInstanceManager.setRef()` so the manager (owned above the boundary, in `InteractionWrapper`) and its plugins re-point at the fresh `Deck` after a remount.
- **`Plot`** owns `onWebGlContextLost` (removed from its public props); caller props are spread **before** the internally owned props so a spread `PlotParams` variable can't overwrite the boundary's handler.
- **`EsvIntersection`** is split: the outer component keeps controller *state* / pub-sub / prop-sync effects / the adapter, while a new inner `EsvIntersectionRenderSurface` owns the container `<div>` and the controller lifecycle and sits *inside* the boundary — so a remount destroys the old controller and rebuilds it against a fresh container instead of orphaning it.

### Notes for reviewers
- The overlay is a plain container — no `role="alert"`. It holds interactive controls (Restore, info popover), so an assertive live region isn't the right fit; announcement semantics can be revisited separately.
- Verified manually: deck.gl readout / picking works both after a normal first mount and after a recovery remount.
- Dashboard-activation was considered as a second auto-recovery trigger but dropped — with a single dashboard, context nullability isn't a real activation signal. Revisit with multi-dashboard.

---

Snapshot for testing: http://localhost:8080/snapshot/frHqydBYsluv


jorgenherje (approved):
Nice work! Well structured and easy to read code!

Tested and it seems to work well.

Also nice with documentation describing the details.

PR #1875, opened 2026-08-28
Fluid contact surface standard result

### LINKED ISSUE #1183: Fluid contacts from the different fluid zones are seen as duplicates
Due to change in metadata => standard_results , fluids contacts do not get uniqueness.
i.e oil and gas contacts are seen as duplicates of the same Sumo object.
https://github.com/equinor/webviz/blob/main/backend_py/primary/primary/services/sumo_access/surface_access.py#L173

We need to update the search context to handle this or start creating separate queries for the different standard results.

The new metadata for this datatype now has a `data.fluid_contact` section that includes the contact name.

HansKallekleiv:
Fluid-contact surfaces are published as the fluid_contact_surface standard result in Sumo. The existing generic surface APIs identify surfaces using name and attribute, which can merge fluid contacts with the same name but different contact types.

Closes #1183 

This PR is the first step toward migrating surface handling to typed standard-result APIs. Eventually, most surfaces are expected to be exposed as standard results with explicit identities and schemas. 
As more standard results is added and we get a better understanding on how the different types look we can start refactoring and abstract the Sumo SurfaceAccess class. For now the current class is expanded with support for fluid contacts.

### Changes
- Added dedicated Sumo queries and API endpoints for fluid-contact surface metadata and data.
- Modelled fluid-contact identity explicitly using surface name and contact type.
- Reused the existing statistical task and long-running operation mechanics.
- Excluded migrated fluid-contact surfaces from the legacy generic surface picker.
- Added a dedicated fluid-contact layer for the 2D Viewer, 3D Viewer, and Intersection module.
- Added readable labels for fluid-contact types.
- Added backend tests for metadata grouping, contact filtering, and statistical queries.

### Notes for reviewers
- Fluid-contact handling currently extends the existing SurfaceAccess class rather than introducing a separate access class or broader abstraction.
- This is intentional. As more surface standard results are added, their shared requirements and differences will become clearer. We can then refactor around established patterns instead of designing an abstraction based on a single standard-result type.
- The implementation still keeps fluid-contact identity and API contracts separate from the legacy generic surface model, while reusing existing retrieval, statistical aggregation, and long-running task mechanics.


sigurdp (approved):
👍 LGTM, as discussed online over Teams.

There are still some minor issues, as discussed, but those will be picked up in a separate PR that gets rid of the current string based handling of standard results.

In addition to that the only minor gripe is that I think the PR message is slightly outdated and should get some love before merging.

